PR #1, opened 2017-11-28
Events / rsvps page

jzhu98:
I've gone through the Django tutorial and pieced together an app to handle the calendar / events section of the website.

There's no real functionality yet (or even a real front-end design yet), but the scaffolding is there to help  with the stupid details - a MySQL connection, a few datatypes for use with the database, routing, front-end templates, and static asset delivery.

AFAIK others should be able to copy and paste the events directory to make their own sections of the website (and register it into hknweb/settings.py and hknweb/urls.py), and Django should be okay with that.

EDIT 11/29: basic HTML layout for events calendar in place. Still puzzling over CSS, but there's a header from a base template (similar to Ryan's header), and the agenda does correctly poll for events from the database. (:O)


hknweb/settings.py
```diff
@@ -75,8 +76,11 @@
 
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        'ENGINE': 'django.db.backends.mysql',
+        'OPTIONS': {
+            'read_default_file': './db.cnf',
```
jvperrin: Nice, this is good. Maybe it would be better to have a directory for config files or something? Anyway, we can worry about that later, it's not like we have too many files in the root directory of the project right now anyway.

hknweb/settings.py
```diff
@@ -75,8 +76,11 @@
...
+            'init_command': "SET sql_mode='STRICT_TRANS_TABLES'",
```
jvperrin: What is this for?

jzhu98:
Was fiddling around with branches in git, and while this merge wasn't intentional it does help to bring pipenv in.

hknweb/settings.py
jzhu98: Django was giving me warnings:
> ?: (mysql.W002) MySQL Strict Mode is not set for database connection 'default'
        HINT: MySQL's Strict Mode fixes many data integrity problems in MySQL, such as data truncation upon insertion, by escalating warnings into errors. It is strongly recommended you activate it. See: https://docs.djangoproject.com/en/2.0/ref/databases/#mysql-sql-mode
So I enabled it.

events/models.py
```diff
@@ -0,0 +1,66 @@
...
+    SEDAN = 4
+    MINIVAN = 6
+
+    TRANSPORT_ENUM = (
+        (NEED_RIDE, "I need a ride"),
+        (HAVE_RIDE, "Don't worry about me"),
+        (SM_SEDAN,  "I have a small sedan (4 seats)"),
+        (SEDAN,     "I have a sedan (5 seats)"),
+        (MINIVAN,   "I have a minivan (7 seats)"),
+    )
```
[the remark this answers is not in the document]
jzhu98: Idk? I was trying to match the current site's schema.

manage.py
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env python
 import os
 import sys
+import pymysql
 
 if __name__ == "__main__":
+    # Expose MySQL interface to Django: https://stackoverflow.com/a/34778155
+    pymysql.install_as_MySQLdb()
```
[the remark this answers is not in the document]
jzhu98: It looks like Django [recommends](https://docs.djangoproject.com/en/2.0/ref/databases/#mysql-db-api-drivers) using [mysqlclient](https://github.com/PyMySQL/mysqlclient-python) as the driver. I'll swap it out.

PR #5, opened 2018-01-29
Add basics of vagrant setup

jvperrin:
This sets up a database and installs a few useful packages. Works pretty well for anyone who doesn't want to install a database on their machine but instead wants a VM option for it.


jzhu98:
Looks good to me; I'll add the pipenv setup into the Vagrantfile shell script.

PR #7, opened 2018-02-22
Events page redux

jzhu98:
Continuation of #1, with most comments addressed:
- Vendor resources (fullcalendar, tachyons, jquery, moment.js) replaced with CDN links ([unpkg.com](https://unpkg.com))
- Removed old Ruby / ERB comments
- PyMySQL replaced with mysqlclient, the Django-recommended mysql / mariadb driver
- Directory structure reorganized, with all apps under the hknweb/ parent

As well as the world's snazziest header / footer nav bars, built with [tachyons](https://tachyons.io).


jvperrin:
(also in the future in general I think we should include screenshots when any pages/things change to make it clear what has been changed and what it looks like, but this is fine for now)

jzhu98:
Pics or it didn't happen:
screenshot from 2018-02-25 11-45-21

PR #8, opened 2018-02-22
Livereload setup

jameslzhu:
Adds a [livereload](http://livereload.com/) setup, invoked with
```sh
$ make livereload
```
which automatically reloads web pages upon detecting source file changes (html, css, js).

(Uses [django-livereload](https://github.com/Fantomas42/django-livereload).)

It was a big help for me when tweaking page designs / css.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,11 @@ dev:
 dev-vagrant:
 	pipenv run python ./manage.py runserver [::]:$(DEV_PORT)
 
+.PHONY: livereload
+livereload:
+	pipenv run python ./manage.py livereload $(DEV_LISTEN_IP):$(DEV_PORT)
+
+
 setup: pipenv venv
 
 .PHONY: pipenv
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,20 +1,20 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3e250b03724612ccadb22ee7e20c5cf9563f8a0752f9374ec0d55b56fa55bebb"
+            "sha256": "572443aad8e855844d94fcecaa649f6f3b99137e429727b121115c695a32f0d6"
         },
         "host-environment-markers": {
             "implementation_name": "cpython",
             "implementation_version": "3.6.4",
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -11,6 +11,11 @@ dev:
 dev-vagrant:
 	pipenv run python ./manage.py runserver [::]:$(DEV_PORT)
 
+.PHONY: livereload
+livereload:
+	pipenv run python ./manage.py livereload $(DEV_LISTEN_IP):$(DEV_PORT)
```
jvperrin: I don't think this'll make it run on the correct port, don't you run this alongside the regular web server to refresh when any changes are made? If so, I think the default port and host it uses are probably fine

Pipfile.lock
```diff
@@ -1,20 +1,20 @@
...
-            "platform_release": "10",
-            "platform_system": "Windows",
-            "platform_version": "10.0.15063",
+            "platform_release": "4.14.20-1-lts",
+            "platform_system": "Linux",
+            "platform_version": "#1 SMP Sat Feb 17 21:35:28 CET 2018",
             "python_full_version": "3.6.4",
             "python_version": "3.6",
-            "sys_platform": "win32"
+            "sys_platform": "linux"
```
jvperrin: Bleh, I don't like how pipenv does all this, I wish it didn't even include this information. Oh well, it's not the worst, just annoying.

Makefile
jameslzhu: Admittedly fuzzy on the details of livereload, but from the cli usage:
```sh
pipenv run python ./manage.py livereload -h
usage: manage.py livereload [-h] [--version] [-v {0,1,2,3}]
                            [--settings SETTINGS] [--pythonpath PYTHONPATH]
                            [--traceback] [--no-color] [-l LIVEPORT]
                            [addrport]

Runs the development server with livereload enabled.

positional arguments:
  addrport              host and optional port the django server should listen
                        on (default: 127.0.0.1:8000)
```
it's the Django port specified, not the livereload server port (at 35729).

PR #9, opened 2018-02-24
Add pre-commit, clean up a few files, and add pytest config

jvperrin:
This uses [pre-commit](https://pre-commit.com/) to automatically test (and sometimes fix!) style issues in the repo. We should set this up with Travis or something to test every commit (along with some actual tests for Django stuff lol), and I think that will make the code a lot more maintainable for the future!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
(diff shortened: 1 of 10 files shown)


jvperrin:
Ok, I added an end to end test with pytest, so that should serve as the start of our testing framework!

PR #10, opened 2018-02-26
Fix a couple broken Vagrant provisioning steps

jvperrin:
`libmariadbclient-dev` isn't in Ubuntu 16.04, so I switched it out for [`libmysqlclient-dev`](https://packages.ubuntu.com/xenial/libmysqlclient-dev), which should work the same essentially. `pip` also isn't available, so I switched it out with `pip3`, and setup needs to be done as the vagrant user to install `pipenv` and the virtualenv to the correct user.

The only change I'm concerned with here is the `pip` -> `pip3` one. Maybe instead the provisioning step can run `pip3 install --user pipenv` and then run `make venv` or something if switching to `pip3` breaks things on windows for example?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ setup: pipenv venv
 
 .PHONY: pipenv
 pipenv:
-	pip install --user pipenv
+	pip3 install --user pipenv
 	@echo Please add $(PIP_HOME) to your PATH variable.
 
 .PHONY: venv
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ setup: pipenv venv
 
 .PHONY: pipenv
 pipenv:
-	pip3 install --user pipenv
+	pip3 install --user pipenv || pip install --user pipenv
 	@echo Please add $(PIP_HOME) to your PATH variable.
 
 .PHONY: venv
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ setup: pipenv venv
 
 .PHONY: pipenv
 pipenv:
-	pip install --user pipenv
+	pip3 install --user pipenv
```
jzhu98: Maybe try pip3, with pip as a fallback on an exit error code?
```
pip3 install --user pipenv || pip install --user pipenv
```
As-is, it probably won't work on Windows. Python 3 installs the binaries as `python.exe` and `pip.exe`. Although the [`py` version selector](https://docs.python.org/3.6/using/windows.html#python-launcher-for-windows) may be helpful, I don't think it's available on Unix.

jzhu98: As an alternative, installing python / pip and python3 / pip3 together will also work without modifying the make rule; pipenv will select Python3 for the venv regardless (as long as it's installed on the host machine).

jvperrin: Ok, sounds good, I probably try the pip3/pip fallback, I think that seems the most flexible to me. I'll try setting up the Vagrant VM I have from scratch and see how that goes.

PR #19, opened 2018-10-26
Setup for development on cloud9, also devel db to SQLite

GeoHutch:
see above


Makefile
```diff
@@ -10,6 +10,10 @@ dev:
 .PHONY: dev-vagrant
 dev-vagrant:
 	HKNWEB_MODE='dev' pipenv run python ./manage.py runserver [::]:$(DEV_PORT)
+
```
jameslzhu: Do you think we could somehow combine all of these `dev-*` using conditional assignment?

Something like

```
IP ?= 127.0.0.1
PORT ?= 3000

.PHONY: dev
dev:
    HKNWEB_MODE='dev' pipenv run python ./manage.py runserver $(IP):$(PORT)
```

and somehow set the Vagrant `IP` env variable to `0.0.0.0`?

hknweb/settings/dev.py
```diff
@@ -6,19 +6,18 @@
...
     from .dummy_secrets import *
 
+import os
+
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = ['localhost','127.0.0.1','hkn.eecs.berkeley.edu','hkn.mu']
 
-INSTALLED_APPS += [
```
jameslzhu: I'm fine with removing livereload, but might be worthwhile to find a way to conditionally enable it with some flag.

hknweb/settings/dev.py
```diff
@@ -6,19 +6,18 @@
...
-    'livereload',
-]
+ALLOWED_HOSTS = ['localhost','127.0.0.1','hkn.eecs.berkely.edu','hkn.mu','hknweb-geohh.c9users.io']
```
jameslzhu: Should be `hkn.eecs.berkeley.edu` (missing 'e').

PR #21, opened 2018-11-17
Officer cleanup

jameslzhu:
Merged officers into the root hknweb application; at some point officership will have to be merged with the django group model, for access permissions.


hknweb/models/officer.py
```diff
@@ -0,0 +1,33 @@
...
+        return "{}@hkn.eecs".format(self.short)
+
+    def __repr__(self):
+        return 'Committee(short={}, name={})'.format(self.short, self.name)
+
+    def __str__(self):
+        return self.name
+
+
+class Officer(models.Model):
```
alexander-zw: Where will we be using this model?

ochan1:
This exists in the code as the Groups model

PR #40, opened 2019-02-11
Make fullcalendar work with database events

Matthew-Signorotti:


hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
...
 class Event(models.Model):
-    id          = models.IntegerField(primary_key=True)
     name        = models.CharField(max_length=255, null=False)
     slug        = models.CharField(max_length=255)
     location    = models.CharField(max_length=255)
     description = models.TextField()
     start_time  = models.DateTimeField(null=False)
     end_time    = models.DateTimeField(null=False)
-    event_type_id = models.IntegerField()
+    rsvp_limit  = models.PositiveIntegerField(null=True, default=999999) # I'm being rather pessimistic in assuming no HKN event will hit 1 million participants!
```
jameslzhu: Since rsvp limits are optional, this might be better encoded as:
```py
rsvp_limit = models.PositiveIntegerField(null=True, blank=True)
```
That way `null=True` means the SQL value can be NULL, and the `blank=True` means in forms this field can be omitted. In particular, when the SQL value is NULL, Event.rsvp_limit will evaluate as `None` in Python (don't quote me, confirmation needed).

Source: https://docs.djangoproject.com/en/2.2/ref/models/fields/#null

hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
...
+    rsvps       = models.IntegerField(default=0)
```
jameslzhu: Perhaps this can also be a PositiveIntegerField?

hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
...
+    # CONFIRMATION = (
+    #     (CONFIRMED,     "Confirmed"),
+    #     (UNCONFIRMED,   "Unconfirmed"),
+    #     (REJECTED,      "Rejected"),
+    # )
+
+    # user  = models.ForeignKey(Profile, on_delete=models.CASCADE)
+    event = models.ForeignKey(Event, on_delete=models.CASCADE)
+    # id    = models.IntegerField(primary_key=True)
+    # confirmed       = models.CharField(max_length=1,
```
jameslzhu: Uh this needs a confirmed field? We want VP / officers to verify candidates showed up.
```py
confirmed = models.BooleanField(default=False)
```

hknweb/events/views.py
```diff
@@ -2,17 +2,35 @@
...
+    return render(request, 'events/show_details.html', context)
+
+
+def rsvp(request, id):
+    # If authenticated and event is under RSVP cap
+    event = Event.objects.get(pk=id)
+    print("rsvps:", event.rsvps)
+    if request.user.is_authenticated and event.rsvps < event.rsvp_limit:
+        event.rsvps += 1
+        event.save()
```
jameslzhu: This may not be an issue, but this may cause a [race condition](https://en.wikipedia.org/wiki/Race_condition): two simultaneous rsvps which both check the `rsvp_limit` may both succeed, but then push the `event.rsvps` beyond `rsvp_limit`.

I'm not sure how to solve this (might be a non-issue, if the rate of responses is low enough), but the technical solution would be either atomic database updates (making sure limit read + rsvp increment cannot be done simultaneously) or transactional updates, below:
https://docs.djangoproject.com/en/2.1/topics/db/transactions/#controlling-transactions-explicitly

hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
...
     slug        = models.CharField(max_length=255)
     location    = models.CharField(max_length=255)
     description = models.TextField()
     start_time  = models.DateTimeField(null=False)
     end_time    = models.DateTimeField(null=False)
-    event_type_id = models.IntegerField()
+    rsvp_limit  = models.PositiveIntegerField(null=True, default=999999) # I'm being rather pessimistic in assuming no HKN event will hit 1 million participants!
+    rsvps       = models.IntegerField(default=0)
+    # event_type_id = models.IntegerField()
     # event_type_id  = models.ForeignKey()
```
jameslzhu: This will tie in with a larger design, but I think we also need an EventType model, with every Event containing a ForeignKey pointing to an EventType. This way candidates will know which event is a Big Fun event, vs. a Prodev or Serv or Fun event. (The idea is that each of these event types is an EventType object stored in the database.)

hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
...
+    # UNCONFIRMED = 'f'
+    # REJECTED = 'r'
+
+    # CONFIRMATION = (
+    #     (CONFIRMED,     "Confirmed"),
+    #     (UNCONFIRMED,   "Unconfirmed"),
+    #     (REJECTED,      "Rejected"),
+    # )
+
+    # user  = models.ForeignKey(Profile, on_delete=models.CASCADE)
```
jameslzhu: I think this is the exact thing you want to tie an RSVP to a specific user.

(I don't know if you want to tie it to Profile or User; I'd go with User.)

hknweb/events/views.py
```diff
@@ -2,17 +2,35 @@
...
+        Rsvp.objects.create(event=event)
+    return HttpResponse("Hello, world. You're at the rsvp index.")
```
jameslzhu: If the user is not authenticated, you should return a 401 Unauthorized or [403 Forbidden](https://docs.djangoproject.com/en/2.1/ref/request-response/#django.http.HttpResponseForbidden) response.

hknweb/events/models.py
```diff
@@ -1,66 +1,53 @@
 from django.db import models
+from hknweb.models import Profile
 
 class Event(models.Model):
-    id          = models.IntegerField(primary_key=True)
     name        = models.CharField(max_length=255, null=False)
     slug        = models.CharField(max_length=255)
```
jameslzhu: There's a dedicated field type, [SlugField](https://docs.djangoproject.com/en/2.1/ref/models/fields/#slugfield), which is meant for url-friendly names like this. (Slug means only alphanumeric + hyphens + underscores). This is meant to be the machine-friendly name to be used in URLs.

hknweb/events/views.py
```diff
@@ -1,18 +1,82 @@
...
+        rsvp = False
+
+    context = {
+        'event': event,
+        'rsvp': rsvp
+    }
+    return render(request, 'events/show_details.html', context)
+
+@csrf_exempt  #doing this for now bc idk how to make csrf work
+def rsvp(request, id):
```
jameslzhu: Apparently there's a `@login_required` decorator you can use for this? 
```py
from django.contrib.auth.decorators import login_required

@login_required
def my_view(request):
    ...
```

https://docs.djangoproject.com/en/2.1/topics/auth/default/#the-login-required-decorator

PR #67, opened 2019-04-28
Shortlinks

shemuelong:
For project (Shem and Rithwik):
Added shortlink url breakage, removed need for /s/ in between shortlinks
Added meta class to change "Linkss" to "Links," but later needs full fix
Changed Links.leader to Links.creator
Added __str__ function for Links display
Added unique field for the names of Links

Updated to newest version of Django 2.2; mainly searched for all deprecated functions (couldn't find any). No other errors seem to be present.


bri25yu:
Stale, work has already been merged by other users

Issue #35, opened 2019-02-02
Preventing bot accounts (captchas)


jameslzhu:
Preventing bot accounts is a concern for the site.

Bot accounts take up username and storage space, and slow down account creation by requiring human intervention (compserv account approval). If access permissions are not done correctly, they may have access to private data.

There are various options we have for this:

- **Do nothing**: Simplest for users, allows all bots. Requires extensive manual intervention.
- **Email filtering**: Allow only emails with *.berkeley.edu domains (in particular, cs.berkeley.edu, ee.berkeley.edu and hkn.eecs.berkeley.edu must also be allowed). This would require email verification (also desirable, perhaps) to avoid being defeated by simply generating a random email.
- [**Google reCAPTCHA v2**](https://www.google.com/recaptcha/intro/v3.html), **currently used**: Offers state-of-the-art bot detection and ease-of-use, but has concerns about privacy (user tracking). That said, we also use Google Analytics anyways, so if we're already making a deal with the devil this will offer significant benefits. Seems to allow some bots through on our current site, though, so it may depend on our settings.
- **Honeypot**: have an invisible form that robots will try to fill out (i.e. CSS `display:none` or `visibility:hidden`), or a form with an obvious human answer (but no obvious robot answer). Can be brute forced if obvious human answer, sometimes bypassed if robots detect invisible form.
- **[Google OAuth signin](https://developers.google.com/identity/sign-in/web/sign-in) (aka OpenID Connect)**. Useful for existing members with @hkn.eecs.berkeley.edu accounts. Enabling this for @berkeley.edu will require interfacing with [CalNet](https://calnetweb.berkeley.edu/calnet-technologists/calnet-terms-service/calnet-authentication-guidelines), which is quite complex.
- **Custom / open-source CAPTCHA**: tend to be difficult to use.
- **Time-based detection**: detecting bots by form population speed (time from load to submit). May be fooled by intelligent bots (false negative) or flag legitimate users with auto-form-fill (false positive).
- **Symbolic math**: generating a random math problem, ideally symbolic to frustrate simple bots, but may require numbers. idk pull some cs70 or cs61a for this, but not everyone's taken cs70.

At the end of it all, there are three major priorities which must be balanced:

1. Preventing bot account creation. This saves compserv time, and keeps data more secure.
2. Keeping human account creation simple. We don't want to frustrate users who can't get past our captcha, which may include those with visual disabilities.
[...]

PR #68, opened 2019-04-28
Google recaptcha + filtering to berkeley emails

briana-jin-zhang:
See issue #35.


hknweb/forms.py
```diff
@@ -21,6 +25,32 @@ class SignupForm(UserCreationForm):
      last_name = forms.CharField(max_length=30, required=True)
      email = forms.EmailField(max_length=200, required=True)
 
+     def clean_email(self):
+         email = self.cleaned_data.get('email')
+         if (email == None or email.split('@')[-1] != 'berkeley.edu'
+         or email.split('@')[1] != email.split('@')[-1]):
+             raise forms.ValidationError('Please use your Berkeley email', code='invalid')
```
michaelmmlu: I think everything looks good but do we only want to allow accounts with Berkeley e-mails?

briana-jin-zhang: I talked to James and he was like yeah, this is the easiest solution rn

PR #89, opened 2019-10-02
Restrict access to candidate portal and improve user admin

alexander-zw:


hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+    def officer(self, user):
+        return 'Y' if user.groups.filter(name=settings.OFFICER_GROUP).exists() else ''
+
+    def candidate(self, user):
+        return 'Y' if user.groups.filter(name=settings.CAND_GROUP).exists() else ''
+
+    def add_cand(self, request, queryset):
+        group = Group.objects.get(name=settings.CAND_GROUP)
+        for u in queryset:
+            group.user_set.add(u)
```
jameslzhu: Consider using
```py
group.user_set.update(queryset)
```
to add all users at once to the group, rather than adding them one at a time.

hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def add_officer(self, request, queryset):
+        group = Group.objects.get(name=settings.OFFICER_GROUP)
+        for u in queryset:
+            group.user_set.add(u)
```
jameslzhu: Ditto.

hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def remove_cand(self, request, queryset):
+        group = Group.objects.get(name=settings.CAND_GROUP)
+        for u in queryset:
+            group.user_set.remove(u)
```
jameslzhu: Ditto.

hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def remove_officer(self, request, queryset):
+        group = Group.objects.get(name=settings.OFFICER_GROUP)
+        for u in queryset:
+            group.user_set.remove(u)
```
jameslzhu: Ditto.

hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    add_cand.short_description = "Add selected as candidates"
```
jameslzhu: Could you move this into each function? (Just to place it next to where it's relevant.)

hknweb/candidate/views.py
```diff
@@ -146,7 +165,11 @@ def send_cand_confirm_email(form):
 
 
 # The page displayed after officer reviews challenge and clicks "submit"
+@login_required(login_url='/accounts/login/')
+@check_account_access
 def officer_review_confirmation(request, pk):
+    check_account_access(request)
```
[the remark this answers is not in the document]
alexander-zw: Oh oops I should have deleted this

hknweb/candidate/views.py
```diff
@@ -158,7 +181,10 @@ def officer_review_confirmation(request, pk):
 
 # Detail view of an officer challenge
 @login_required(login_url='/accounts/login/')
+@check_account_access
 def challenge_detail_view(request, pk):
+    check_account_access(request)
```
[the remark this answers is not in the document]
alexander-zw: Wow

hknweb/candidate/views.py
```diff
@@ -4,18 +4,34 @@
 from django.core.mail import EmailMultiAlternatives
 from django.template.loader import render_to_string
 from django.contrib import messages
-from django.contrib.auth.decorators import login_required
+from django.contrib.auth.decorators import login_required, user_passes_test
 from django.utils.decorators import method_decorator
 from django.conf import settings
 from django.contrib.staticfiles.finders import find
 from django.contrib.staticfiles.templatetags.staticfiles import static
+from django.conf import settings
```
jameslzhu: This is a duplicated import?

hknweb/admin.py
```diff
@@ -0,0 +1,47 @@
...
+@admin.register(User)
+class CustomUserAdmin(UserAdmin):
+    list_display = ('username', 'first_name', 'last_name', 'email', 'officer', 'candidate')
+    readonly_fields = ('officer', 'candidate')
+
+    actions = ['add_cand', 'add_officer', 'remove_cand', 'remove_officer']
+
+    def officer(self, user):
+        return 'Y' if user.groups.filter(name=settings.OFFICER_GROUP).exists() else ''
+
```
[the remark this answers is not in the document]
alexander-zw: If I return True and False, the admin page will display "True" or "False" instead of checkmarks and Xs, which is hard to read. I don't know how to get it to display the checkmarks; it only does that for boolean fields, not functions returning booleans. So 'Y' and '' is the most readable I can come up with.

hknweb/candidate/views.py
```diff
@@ -4,18 +4,34 @@
 from django.core.mail import EmailMultiAlternatives
 from django.template.loader import render_to_string
 from django.contrib import messages
-from django.contrib.auth.decorators import login_required
+from django.contrib.auth.decorators import login_required, user_passes_test
```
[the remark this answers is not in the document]
alexander-zw: Oh I guess I forgot to remove that

hknweb/candidate/views.py
```diff
@@ -4,18 +4,34 @@
...
+            return render(request, "errors/401.html", status=401)
+        return func(request, *args, **kwargs)
+    return check_then_call
+
+
+# views
+
 # Candidate portal home
 @method_decorator(login_required(login_url='/accounts/login/'), name='dispatch')
+@method_decorator(check_account_access, name='dispatch')
```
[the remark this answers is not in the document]
alexander-zw: The dispatch function is inherited from the parent class. I used method decorators because that's what the example I found did, and it also that way you can use the same decorator for methods and functions.

hknweb/admin.py
alexander-zw: How do you know all these things

hknweb/admin.py
alexander-zw: Inside the function? Would it work?

jameslzhu:
@alexander-zw https://docs.djangoproject.com/en/2.2/ref/contrib/auth/#group-model
I'm noticing none of the groups are showing up in the admin console; they are actual models, if instead of having string constants in settings we do a Groups.find(name="candidates") that is a viable alternative

hknweb/admin.py
jameslzhu: Denero did it; if not inside the function then maybe next to it

jameslzhu: (This is entirely stylistic)

hknweb/admin.py
jameslzhu: I figured there would be some function to handle iterables instead of just single elements, the way lists have .append() and .extend(), and the rest was searching the Django queryset docs.

alexander-zw: Uh actually this is more complicated so I'll just keep it

hknweb/admin.py
alexander-zw: I still kinda like this way

alexander-zw: It didn't work putting it in the function, I'll add it right after each function

hknweb/candidate/views.py
alexander-zw: Yes

hknweb/candidate/views.py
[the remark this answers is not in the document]
jameslzhu: Could you post a link to the example?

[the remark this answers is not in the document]
jameslzhu: It looks like there are less-verbose alternatives to these decorators, namely mixin classes to inherit from: https://docs.djangoproject.com/en/2.2/topics/auth/default/#django.contrib.auth.mixins.LoginRequiredMixin

I can rewrite this code with mixin inheritance in a separate pull request.

PR #93, opened 2019-10-03
Upcoming events feature

whuang7000:
Lists all Event objects within the next week


hknweb/candidate/views.py
```diff
@@ -10,8 +10,10 @@
 from django.contrib.staticfiles.finders import find
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from random import randint
+import datetime
 
 from .models import OffChallenge, Announcement
+from ..events.models import Event
```
jameslzhu: Why the double dot?

hknweb/candidate/templates/candidate/index.html
```diff
@@ -120,7 +120,27 @@ <h4>Bit-Byte</h4>
...
+                    <th>Event</th>
+                    <th>Start Time</th>
+                    <th>End Time</th>
+                    <th>Location</th>
+                </thead>
+
+                <tbody>
+                {% for event in upcoming_events %}
+                    <tr>
+                        <td align="center"> {{ event.name }} </td>
```
jameslzhu: Did you mean style=“text-align:center” instead of align=“center”?

As a side note, if it’s a table of names, left aligning usually looks better than center aligning.

hknweb/candidate/views.py
```diff
@@ -28,11 +30,14 @@ def get_context_data(self):
...
                 .order_by('-release_date')
+        today = datetime.date.today()
+        upcoming_events = Event.objects.filter(start_time__range=(today, today + datetime.timedelta(days=7))).order_by('start_time')
         context = {
             'num_pending' : challenges.filter(reviewed=False).count(),
             'num_rejected' : reviewed_challenges.filter(confirmed=False).count(),
             'num_confirmed' : reviewed_challenges.filter(confirmed=True).count(),
-            'announcements' : announcements
+            'announcements' : announcements,
+            'upcoming_events': upcoming_events
```
jameslzhu: I tend to put trailing commas to minimize future diffs:

‘upcoming_events’: upcoming_events,

hknweb/candidate/views.py
jameslzhu: Oh, it’s from the events in the parent folder...
Uh, looks a little weird but I guess it works.

hknweb/candidate/views.py
whuang7000: Fixed

hknweb/candidate/templates/candidate/index.html
whuang7000: Fixed

jameslzhu:
As a note, it looks like the align=left is the default, and furthermore is [deprecated in HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td).

I’d recommend you just remove the attribute from <td>.

PR #95, opened 2019-10-04
Fix RSVP logic and remove rsvps field from RSVP model

whuang7000:
RSVP logic wasn't working properly, I removed the "rsvps" field from the RSVP model because it is an unreliable way to check the number of rsvps. Instead, I just used the length of the foreignkey queryset. Also added a null check to rsvp_limit in the logic.


hknweb/events/views.py
```diff
@@ -33,8 +33,9 @@ def rsvp(request, id):
         raise Http404()
 
     event = Event.objects.get(pk=id)
-    if request.user.is_authenticated and event.rsvps < event.rsvp_limit:
-        event.rsvps = F("rsvps") + 1
+    rsvps = len(event.rsvp_set.all())
```
jameslzhu: Try `event.rsvp_set.count()` instead?
https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.count

hknweb/events/views.py
```diff
@@ -33,8 +33,9 @@ def rsvp(request, id):
...
+
+    if request.user.is_authenticated or not event.rsvp_limit and rsvps < event.rsvp_limit:
```
jameslzhu: extraordinarily ambiguous (operator precedence), pls parenthesize

hknweb/events/views.py
```diff
@@ -47,9 +48,7 @@ def unrsvp(request, id):
         raise Http404()
 
     event = Event.objects.get(pk=id)
-    if request.user.is_authenticated and event.rsvps < event.rsvp_limit:
-        #check if rsvp for this event and this user already exists; if false, then set true
-        event.rsvps = F("rsvps") - 1
+    if request.user.is_authenticated:
         Rsvp.objects.get(user=request.user, event=event).delete()
```
jameslzhu: This will throw a [`Model.DoesNotExist` error](https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.DoesNotExist) if the Rsvp object does not exist; perhaps handle this exception?

PR #99, opened 2019-10-04
Fixed event creation and authentication

chevin-ken:


hknweb/events/views.py
```diff
@@ -1,20 +1,22 @@
...
 from hknweb.models import Profile
 
 def index(request):
     events = Event.objects.order_by('-start_time')
+    
     context = {
         'events': events,
     }
-    return render(request, 'events/index.html', context)
+    return render(request, 'events/index.html', context, RequestContext(request, context))
```
jameslzhu: eh. probably `RequestContext(request, {})`?

Issue #94, opened 2019-10-03
Add data migration for initial (seed) groups in database


jameslzhu:
https://docs.djangoproject.com/en/2.2/topics/migrations/#data-migrations

We'll need to add the following Groups for permissions to work:

- [ ] officer
- [ ] candidate

(not plural)

This is required for new development instances (i.e. new developers), who initialize new (blank) databases and thus will have any functionality involving candidates or officers broken until they manually add candidate and officer groups.

bri25yu:
Closed by https://github.com/compserv/hknweb/pull/238

PR #111, opened 2019-10-09
Events authentication and other small changes

alexander-zw:
Redirect user to login when they click on an event in the calendar


hknweb/events/models.py
```diff
@@ -36,9 +36,9 @@ def __str__(self):
        return self.name
 
 class Rsvp(models.Model): # TODO: null should be false in some cases
-    user  = models.ForeignKey(User, models.CASCADE, null=True, verbose_name="rsvp'd by")
-    event = models.ForeignKey(Event, models.CASCADE, null=True)
-    confirmed       = models.BooleanField(null=True)
+    user  = models.ForeignKey(User, models.CASCADE, null=False, verbose_name="rsvp'd by")
+    event = models.ForeignKey(Event, models.CASCADE, null=False)
+    confirmed       = models.BooleanField(null=False, default=False)
```
jameslzhu: null=False is actually the default value, fyi.

hknweb/events/views.py
```diff
@@ -41,27 +62,34 @@ def rsvp(request, id):
...
         raise Http404()
 
     event = get_object_or_404(Event, pk=id)
 
     if request.user.is_authenticated:
-        rsvp = get_object_or_404(Rsvp, user=request.user, event=event)
-        rsvp.delete()
-        messages.success(request, 'un-RSVP\'d :(')
+        # check if rsvp for this event and this user already exists; if false, then set true
+        if datetime.datetime.now().replace(tzinfo=pytz.UTC) > event.end_time.replace(tzinfo=pytz.UTC):
```
jameslzhu: https://docs.djangoproject.com/en/2.2/topics/i18n/timezones/

The better timezone-aware strategy would be to use the Django timezone import:
```
now = timezone.now()
```

hknweb/events/views.py
```diff
@@ -41,27 +62,34 @@ def rsvp(request, id):
...
+            messages.error(request, 'Could not un-RSVP: the event has already ended.')
+        else:
+            rsvp = get_object_or_404(Rsvp, user=request.user, event=event)
+            rsvp.delete()
+            messages.success(request, 'un-RSVP\'d :(')
     else:
         messages.error(request, 'Something went wrong; could not un-RSVP.')
     return redirect('/events/' + str(id))
```
jameslzhu: There might be a better way of doing this with [`get_absolute_url()`](https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.get_absolute_url) but we can address this separately.

https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#redirect

hknweb/events/views.py
```diff
@@ -4,13 +4,32 @@
 from django.contrib import messages
 from django.shortcuts import get_object_or_404
 from django.views.decorators.csrf import csrf_exempt #doing this for now bc idk how to make csrf work
-from django.db.models import F # Avoids changing database values without risking a race condition
+from django.conf import settings
 from django.contrib.auth.decorators import login_required, permission_required
+import datetime
+import pytz
```
jameslzhu: Given the below changes to use `from django.utils import timezone`, this would make these two imports (`datetime`, `pytz`) unnecessary.

hknweb/events/views.py
```diff
@@ -28,11 +49,11 @@ def show_details(request, id):
...
         'rsvp': rsvp,
     }
-    if request.user.is_authenticated:
-        return render(request, 'events/show_details.html', context)
-    return redirect('/events')
+    return render(request, 'events/show_details.html', context)
 
 @csrf_exempt  #doing this for now bc idk how to make csrf work
+@login_required(login_url='/accounts/login/')
+@check_account_access
```
jameslzhu: How would you feel about using a more granular permission that can be granted to multiple groups, like:
```py
from django.contrib.auth.decorators import permission_required

@permission_required('events.add_rsvp', login_url='/accounts/login')
def my_view(request):
    ...
```
https://docs.djangoproject.com/en/2.2/topics/auth/default/#the-permission-required-decorator

Each model ('Bar') has the following [default permissions](https://docs.djangoproject.com/en/2.2/topics/auth/default/#default-permissions) generated:

```
add: user.has_perm('foo.add_bar')
change: user.has_perm('foo.change_bar')
delete: user.has_perm('foo.delete_bar')
view: user.has_perm('foo.view_bar')
```

This would have the benefit of not having to add groups to `@check_account_access` as we add groups, as long as we grant this permission to each group.
As a side effect, this requires only one decorator instead of two.

Also, the [`settings.LOGIN_URL`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-LOGIN_URL) appears to set the value of `login_url`, which furthermore has default value `/accounts/login/`, so this parameter is actually unnecessary in our project:

```py
@permission_required('events.add_rsvp')
def my_view(request):
    ...
```

hknweb/events/views.py
```diff
@@ -4,13 +4,32 @@
...
+def check_account_access(func):
+    def check_then_call(request, *args, **kwargs):
+        if not is_cand_or_officer(request.user):
+            return render(request, "errors/401.html", status=401)
+        return func(request, *args, **kwargs)
+    return check_then_call
+
+def is_cand_or_officer(user):
+    return user.groups.filter(name=settings.CAND_GROUP).exists() or \
+           user.groups.filter(name=settings.OFFICER_GROUP).exists()
```
jameslzhu: See below comment on suggestion to switch to fine-grained permission access control over coarse group access control.

hknweb/events/views.py
```diff
@@ -19,6 +38,8 @@ def index(request):
     }
     return render(request, 'events/index.html', context, RequestContext(request, context))
 
+@login_required(login_url='/accounts/login/')
+@check_account_access
```
jameslzhu: See above.

hknweb/events/views.py
```diff
@@ -41,27 +62,34 @@ def rsvp(request, id):
...
-        Rsvp.objects.create(user=request.user, event=event)
+        Rsvp.objects.create(user=request.user, event=event, confirmed=False)
         messages.success(request, 'RSVP\'d!')
     else:
         messages.error(request, 'Could not RSVP; the RSVP limit has been reached.')
     return redirect('/events/' + str(id))
 
 @csrf_exempt  #doing this for now bc idk how to make csrf work
+@login_required(login_url='/accounts/login/')
+@check_account_access
```
jameslzhu: See above.

hknweb/events/models.py
alexander-zw: I know but we can be explicit

hknweb/events/views.py
alexander-zw: LOL that ridiculously outdated comment is still here

hknweb/events/views.py
alexander-zw: If we did this, we would need to add the appropriate permissions to the candidate and officer group, which is easier to organize, but more work to set up

alexander-zw: Is there a way to indicate what permissions each group should have, and even automatically add them upon startup or something

jameslzhu: The way to do that would be a data migration, listed in https://github.com/compserv/hknweb/issues/94:

https://docs.djangoproject.com/en/2.2/topics/migrations/#data-migrations

This would allow us to create a database migration that adds data, instead of modifying the tables.

hknweb/events/models.py
jameslzhu: Explicit is (usually) good, but here we're not even consistent with it; we omit it from literally every other field and it's the default.

I feel like consistency trumps explicitness here; on reading this code, usually the (naive) assumption is that the default is the opposite of the explicit value given.

hknweb/events/views.py
alexander-zw: How about I merge this for now, then create a separate pull request to refactor this?

PR #114, opened 2019-10-09
Fixed improper time zone display

chevin-ken:
I didn't delete a lot of the code in landing.py even though it is commented out because I want to refer to it for a better solution in the future that handles the time zone logic without using javascript scripts during the page rendering


hknweb/views/landing.py
```diff
@@ -11,29 +11,29 @@ def home(request):
...
+    #     return {'day': day, 'start': start_time, 'end': end_time}
+    # upcoming_events = Event.objects.filter(start_time__gte=today).order_by('start_time')
+    # if len(upcoming_events) > 4:
+    #     upcoming_events = upcoming_events[:4]
+    # events = []
+    # for event in upcoming_events:
+    #     temp_dict = parse_date(event)
+    #     temp_dict['name'] = event.name
+    #     events.append(temp_dict)
+    events = Event.objects.filter(start_time__gte=today).order_by('start_time')
```
jameslzhu: I'm getting a warning while running this that `today` is a naive datetime:
```
/home/jzhu/.local/share/virtualenvs/hknweb-jdaoxjVp/lib/python3.7/site-packages/django/db/models/fields/__init__.py:1421: RuntimeWarning: DateTimeField Event.start_time received a naive datetime (2019-10-10 22:04:06.644268) while time zone support is active.
```
I think this is because `time` is a `datetime` object without explicit timezone support. The better alternative is to use 
```

now = timezone.now()
```

PR #116, opened 2019-10-13
Add Officer portal to nav bar and change wording: profile -> candidate portal

alexander-zw:


hknweb/settings/common.py
```diff
@@ -78,6 +78,8 @@
                 'django.contrib.messages.context_processors.messages',
                 'social_django.context_processors.backends',
                 'social_django.context_processors.login_redirect',
+                'social_django.context_processors.login_redirect',
```
Matthew-Signorotti: Why is the same line twice?

hknweb/candidate/views.py
```diff
@@ -19,6 +19,7 @@
 # decorators
 
 # used for things only officers and candidates can access
+# TODO: use permissions instead of just the groups
```
Matthew-Signorotti: If this is still a to-do, should we just do it this week before merging this pull request?

hknweb/settings/common.py
```diff
@@ -78,6 +78,8 @@
...
+                'hknweb.views.users.add_officer_context',
```
Matthew-Signorotti: So let me get this straight...for some reason, these functions are called as "context processors" which allow variables like `viewer_is_an_officer` to be referenced in templates like `base.html`? Sounds pretty confusing, but I'm not sure of a better solution. Maybe ask @jameslzhu on this one since I've never dealt with context processors before.

hknweb/templates/base.html
```diff
@@ -125,7 +125,11 @@
     </div>
     {% if user.is_authenticated %}
     <div id="account-nav-bar" class="sub-nav-bar navi-container">
-      <a href="/cand">Profile</a>
+      {% if viewer_is_an_officer %}
```
Matthew-Signorotti: This all seems fine on the front-end! The only thing I'm questioning is how we're determining who is an officer.

hknweb/settings/common.py
alexander-zw: Uh I commented OOPS on this line but apparently forgot to publish it

hknweb/candidate/views.py
alexander-zw: This is for the future

hknweb/settings/common.py
alexander-zw: The other possible way is to use template tags, but that seems to require creating a new folder and two new files, just for one function

hknweb/templates/base.html
alexander-zw: It checks whether they're in the officer group

PR #118, opened 2019-10-14
Dynamic candidate checklist

dinganthony:
- Checklist now has icons for each requirement that change based on whether candidates completed them or not
- All confirmed (green bullet) and unconfirmed (red bullet) RSVPs listed under their requirement
- CSS is pretty basic at the moment and the requirement event types are hard coded
- Still need to do officer hangouts and bitbyte, which should be handled separately

### DIFF
--- a/hknweb/candidate/templates/candidate/index.html
+++ b/hknweb/candidate/templates/candidate/index.html
@@ -72,21 +72,26 @@ Candidate Portal
 
             <div class="checklist-left-col">
                 <ul class="checkboxes" style="margin-bottom: 0;">
+                    <br/>
                     <li>
-                    {% if req_statuses.mandatory_meetings %}
-                        <span class="req done"></span>
-                    {% else %}
-                        <span class="req not-done"></span>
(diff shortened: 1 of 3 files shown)


hknweb/candidate/templates/candidate/index.html
```diff
@@ -71,53 +71,122 @@ <h2>Candidate Requirements</h2>
...
-                        <li>Thing 2</li>
-                        <li>Thing 3</li>
-                    </ol>
-                    <li>Big Fun</li>
-                    <li>Service</li>
-                    <li>Prodev</li>
+                <ul class="checkboxes" style="margin-bottom: 0;">
+                    <br/>
+                    <li>
+                        {% if req_statuses.mandatory_meetings %}
```
alexander-zw: Could you make this a loop over event times instead of doing each one?

alexander-zw: Never mind, if we use a loop it won't preserve order

PR #127, opened 2019-10-26
Challenge fixes

alexander-zw:
Add csec confirmation and full name of officers next to username


hknweb/candidate/views.py
```diff
@@ -255,7 +279,7 @@ def sort_rsvps_into_events(rsvps):
 
 # Checks which requirements have been fulfilled by a candidate
 def check_requirements(sorted_rsvps):
-    #TODO: increase flexibility by fetching event requirement count from database
+    # TODO: increase flexibility by fetching event requirement count from database
```
jameslzhu: Probably be easier to track TODOs like these if you file issues for these instead of waiting for someone to grep for these.

hknweb/candidate/admin.py
```diff
@@ -1,26 +1,44 @@
...
 
+    def save_model(self, request, obj, form, change):
+        super().save_model(request, obj, form, change)
+        if 'csec_confirmed' in form.changed_data:
+            OffChallengeAdmin.check_send_email(request, obj)
+
+    @staticmethod
+    def check_send_email(request, obj):
+        # officer has already confirmed, and now csec confirms
+        if obj.csec_confirmed is True and obj.officer_confirmed is True:
```
jameslzhu: `is True` is rather unnecessary, and readability is preserved without it:
```py
if obj.csec_confirmed and obj.officer_confirmed
```
In particular, None is falsey.

hknweb/candidate/admin.py
```diff
@@ -1,26 +1,44 @@
...
+            send_cand_confirm_email(request, obj, True)
+        # officer has not already rejected, and now csec rejects
+        elif obj.csec_confirmed is False and obj.officer_confirmed is not False:
```
jameslzhu: Does it really matter how the officer has responded if csec has rejected?

hknweb/candidate/admin.py
```diff
@@ -32,12 +50,30 @@ def export_as_csv(self, request, queryset):
...
+                OffChallengeAdmin.check_send_email(request, obj)
+
+    csec_confirm.short_description = "Mark selected as confirmed (csec)"
+
+    def csec_reject(self, request, queryset):
+        for obj in queryset:
+            if obj.csec_confirmed is not False:
+                obj.csec_confirmed = False
+                obj.save()
+                OffChallengeAdmin.check_send_email(request, obj)
```
jameslzhu: This should be `self.check_send_email`, because even if it's a static method we want this code to survive class renaming.

hknweb/candidate/admin.py
```diff
@@ -32,12 +50,30 @@ def export_as_csv(self, request, queryset):
...
         return response
 
     export_as_csv.short_description = "Export selected as csv"
 
+    def csec_confirm(self, request, queryset):
+        for obj in queryset:
+            if obj.csec_confirmed is not True:
+                obj.csec_confirmed = True
+                obj.save()
+                OffChallengeAdmin.check_send_email(request, obj)
```
jameslzhu: Ditto.

hknweb/candidate/models.py
```diff
@@ -13,28 +12,34 @@ class OffChallenge(models.Model):
...
-                        on_delete=models.CASCADE, default=None, related_name='requester')
-    officer         = models.ForeignKey('auth.User', limit_choices_to={'groups__name': settings.OFFICER_GROUP},
-                        on_delete=models.CASCADE, default=None, related_name='officer')
-    name            = models.CharField(max_length=MAX_STRLEN, default='', verbose_name="title")
-    description     = models.TextField(max_length=MAX_TXTLEN, blank=True, default='')
+    # the requester needs permission to add officer challenges. This limits the choices in a dropdown,
+    # but the view CandRequestView is still able to set requesters regardless of their permissions
+    requester         = models.ForeignKey('auth.User',
+    # NOTE: for some reason this causes an error when you try to save on the admin, so I'll just comment it out for now
+                            # limit_choices_to=Q(groups__permissions__codename='add_offchallenge'),
```
jameslzhu: I'll look into how to do this.

hknweb/candidate/templates/candidate/challenge_detail.html
```diff
@@ -19,23 +19,23 @@
...
         <img src="{% static 'candidate/yes.png'%}" alt="confirmed" class="inline-icon">
-        Confirmed uwu
+        Confirmed uwu</br>
     {% else %}
-        <img src="{% static 'candidate/no.png'%}" alt="declined" class="inline-icon">
-        Declined :<</b>
-    {% endif %}<br/>
+        <img src="{% static 'candidate/maybe.png'%}" alt="in progress" class="inline-icon">
+        Waiting for officer/csec response</br>
+    {% endif %}
```
jameslzhu: Do you think it would be better to have two checkmarks, for officer / csec confirmation, or have one status (at the cost of code complexity)?

hknweb/candidate/templates/candidate/challenge_detail.html
```diff
@@ -19,23 +19,23 @@
...
-        <img src="{% static 'candidate/maybe.png'%}" alt="in progress" class="inline-icon">
-        Waiting for officer response
-    {% elif challenge.confirmed %}
+    {% if challenge.officer_confirmed is False or challenge.csec_confirmed is False %}
+        <img src="{% static 'candidate/no.png'%}" alt="declined" class="inline-icon">
+        Declined :<</br>
+    {% elif challenge.officer_confirmed is True and challenge.csec_confirmed is True %}
         <img src="{% static 'candidate/yes.png'%}" alt="confirmed" class="inline-icon">
-        Confirmed uwu
+        Confirmed uwu</br>
```
jameslzhu: -_-

hknweb/candidate/templates/candidate/offreq.html
```diff
@@ -36,12 +37,21 @@ <h3>Past Officer Challenges</h3>
                         </td>
                         <td align="center">{{ challenge.requester }}</td>
                         <td align="center">
-                            {% if not challenge.reviewed %}
-                                <img src="{% static 'candidate/maybe.png'%}" alt="in progress" class="table-icon">
-                            {% elif challenge.confirmed %}
+                            {% if challenge.officer_confirmed is False or challenge.csec_confirmed is False %}
```
jameslzhu: If we're using this kind of conditional logic repeatedly, maybe we should have methods on the OffChallenge object for confirmed() or declined() or in_progress().

hknweb/candidate/templates/candidate/offreq.html
```diff
@@ -60,6 +70,10 @@ <h3>Past Officer Challenges</h3>
         Waiting for response <img src="{% static 'candidate/maybe.png'%}" alt="in progress" class="inline-icon">
         </p>
 
+        <p>
+            Note: Both you and csec need to confirm a challenge for it to be valid.
```
jameslzhu: Is the officer seeing this or the candidate? If it's the candidate, this needs to be reworded to "Both your officer and csec."

hknweb/candidate/views.py
```diff
@@ -41,21 +42,33 @@ def get_context_data(self):
...
+                .count()
+        num_rejected = challenges \
+                .filter(Q(officer_confirmed=False) | Q(csec_confirmed=False)) \
+                .count()
+
         announcements = Announcement.objects \
                 .filter(visible=True) \
                 .order_by('-release_date')
+
         today = datetime.date.today()
```
jameslzhu: Try replacing this with the timezone-aware equivalent:

```

today = timezone.now()
```

hknweb/candidate/views.py
```diff
@@ -163,15 +158,22 @@ def send_cand_confirm_email(form):
...
 
     if form.is_valid():
         form.instance.reviewed = True
         form.save()
-        send_cand_confirm_email(form)
+        # csec has already confirmed, and now officer confirms
+        if challenge.officer_confirmed is True and challenge.csec_confirmed is True:
+            send_cand_confirm_email(request, form.instance, True)
+        # csec has not already rejected, and now officer rejects
+        elif challenge.officer_confirmed is False and challenge.csec_confirmed is not False:
```
jameslzhu: Another good place to reuse the confirmed() and rejected() methods I suggested earlier.

hknweb/candidate/views.py
```diff
@@ -234,10 +236,32 @@ def get_rand_photo(width=400):
         urls = f.readlines()
     return urls[randint(0, len(urls) - 1)].strip() + "?w=" + str(width)
 
-#Takes in all confirmed rsvps and sorts them into types, current hard coded
+def send_cand_confirm_email(request, challenge, confirmed):
+    subject = 'Your Officer Challenge Was Reviewed'
```
jameslzhu: I Have Opinions About The Use Of Capitalization In Email Titles. This Feels Too Long To Capitalize, I Think
```
subject = '[HKN] Your officer challenge was reviewed!'
```
looks better.

hknweb/candidate/views.py
```diff
@@ -234,10 +236,32 @@ def get_rand_photo(width=400):
...
+    candidate_email = challenge.requester.email
+    text_content = 'Your Officer Challenge Was Reviewed'
```
jameslzhu: Ditto.

hknweb/candidate/views.py
alexander-zw: There are so many issues... though is one in particular does seem like a long term investment

alexander-zw: What if we have both? An issue so we know it's there, and a TODO so you know where in the code to look. Can't hurt

hknweb/candidate/admin.py
alexander-zw: I find this more intuitive since this is more like a three state enum rather than a boolean

hknweb/candidate/admin.py
alexander-zw: We don't want to send two emails for the same thing do we?

hknweb/candidate/admin.py
alexander-zw: Huh I never heard that

alexander-zw: Isn't there a thing called refactor?

hknweb/candidate/templates/candidate/offreq.html
alexander-zw: It's the officer

hknweb/candidate/admin.py
jameslzhu: ... how do we send two emails? There's no email sent here for if the officer has rejected.

hknweb/candidate/admin.py
jameslzhu: An IDE feature, perhaps, but depending on it for a team with inconsistent editors and IDEs will require more effort than saved.

hknweb/candidate/views.py
jameslzhu: So issues can reference where to look in the code, too: https://github.com/compserv/hknweb/blob/0306ddcb3a6fa7ae3c984dc391eb8a9a7e5d2738/hknweb/candidate/views.py#L282

jameslzhu: But ok, fine, you can leave the TODO as long as it's accompanied by an issue.

hknweb/candidate/admin.py
alexander-zw: If the officer already rejected, an email was already sent. So you don't need to send another one

hknweb/candidate/templates/candidate/challenge_detail.html
alexander-zw: The candidate doesn't need to know the details, but the officer needs to know

hknweb/candidate/templates/candidate/offreq.html
alexander-zw: Can you access those methods from Django templates

alexander-zw: I used a property instead! Unfortunately I can't use properties to filter querysets, so index.html still duplicates code

hknweb/candidate/views.py
alexander-zw: I didn't even write this code though

alexander-zw: While I'm at it I'll just make even more changes

hknweb/candidate/views.py
alexander-zw: So the other subject is fine? 'Confirm Officer Challenge'

hknweb/candidate/views.py
alexander-zw: This is slightly different though, you're checking if it was previously confirmed, and if it's now confirmed

PR #128, opened 2019-10-26
Add more info in event detail page along with link to go back to cale…

alexander-zw:
…ndar


alexander-zw:
I already merged this, but we might want to have a short review anyway

PR #129, opened 2019-10-26
Add the day of the week in event detail

alexander-zw:
(See other pull request)


jameslzhu:
Should we combine this with #130? It's a much cleaner design.

alexander-zw:
Yes, I don't know how to combine them

jameslzhu:
I can combine them; if you agree we can close this pull request and I'll implement in #130.

PR #130, opened 2019-10-27
Datetime cleanup (home / event detail)

jameslzhu:
https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#date

This pull request simplifies the datetime parsing code by using format strings for the `date` and `time` template filters (linked above):
```html
<p><b>Date</b>: {{event.start_time|date:"D, N j, Y"}}</p>
<p><b>Time</b>: {{event.start_time|time}} to {{event.end_time|time}}</p>
...
<span class="events-card-time">{{event.start_time|date:"D m/d h:i A"}} - {{ event.end_time|date:"h:i A"}}</span>
```

It also changes the events shown on the front page to those *ending* after now(), not just those *starting* after now; viewing in-progress events is useful especially if you're late.

PR #134, opened 2019-10-28
Show rsvps in event detail page

alexander-zw:


hknweb/candidate/templates/candidate/candreq.html
```diff
@@ -57,7 +57,8 @@ <h3>Request Officer Challenge Confirmation</h3>
     </div>
 
     <!-- Right side of page, list of requested challenges -->
-    <div id='agenda'  style="width: 50%; float: right; padding-left: 20px; padding-right: 20px;">
+    <!--  TODO: refactor CSS (e.g. move into style.css and change px into em)  -->
+    <div id='agenda' style="width: 50%; float: right; padding-left: 20px; padding-right: 20px;">
```
jameslzhu: Is now a good time to do this todo? I imagine it'd look like
```css
#agenda {
    width: 50%;
    float: right;
    padding-left: 1.5em;
    padding-right: 1.5em;
}
```

But I'm not exactly sure what the em values are.

hknweb/events/templates/events/add_event.html
```diff
@@ -3,6 +3,7 @@
 {% block title %}Add an event{% endblock %}
 
 {% block header %}
+<link rel="stylesheet" href="{% static "events/style.css" %}">
 <style>
```
jameslzhu: Why would you move only part of the CSS into a style.css?

hknweb/events/templates/events/show_details.html
```diff
@@ -3,63 +3,101 @@
...
+      if (sameDayEvent) {
+        var HTMLtext = '<p><b>Date</b>: ' + start.toLocaleDateString() +
+                       '</p> <p><b>Time</b>: ' + start.toLocaleTimeString([], {hour: '2-digit', minute:'2-digit'}) + ' to ' +
+                       end.toLocaleTimeString([], {hour: '2-digit', minute:'2-digit'});
+      } else { // Event starts and ends on different days
+        var HTMLtext = '<p><b>Starts</b>: ' + start.toLocaleDateString() + ' at ' +
+                       start.toLocaleTimeString([], {hour: '2-digit', minute:'2-digit'}) + '</p> <p><b>Ends</b>: ' +
+                       end.toLocaleDateString() + ' at ' + end.toLocaleTimeString([], {hour: '2-digit', minute:'2-digit'});
+      }
+      document.getElementById("date-and-time").innerHTML = HTMLtext;
```
jameslzhu: How do you think we should reconcile this with #130? Merge #130 first and / or merge its changes into this pull request?

hknweb/events/views.py
```diff
@@ -42,10 +42,12 @@ def show_details(request, id):
...
     event = get_object_or_404(Event, pk=id)
 
-    rsvp = Rsvp.objects.filter(user=request.user, event=event).exists()
+    rsvpd = Rsvp.objects.filter(user=request.user, event=event).exists()
+    other_rsvps = Rsvp.objects.filter(event=event).exclude(user=request.user)
     context = {
         'event': event,
-        'rsvp': rsvp,
+        'rsvpd': rsvpd,
+        'other_rsvps': other_rsvps,
```
jameslzhu: Beyond the nifty (oh look I rsvp'd) reason of listing "You" as an rsvp, is there any other reason to not explicitly list the person's name / email (and include the request.user in the query)?
I imagine if csec wants a record of everyone at an event (or any other officer), they need to copy + paste the list and then add themselves, esp. for reimbursements.

hknweb/candidate/templates/candidate/candreq.html
alexander-zw: Better than using px right?

hknweb/events/templates/events/add_event.html
alexander-zw: Because some of it is specific to the html, I moved out as much as I could

alexander-zw: Because some of it is specific to the html, I moved out as much as I could

hknweb/events/views.py
alexander-zw: Csec could user the admin page

hknweb/events/templates/events/show_details.html
chevin-ken: Probably merge #130 into this

hknweb/events/views.py
jameslzhu: csec, perhaps, but other officers? So far we haven't given most officers access to the admin page, nor have we decided to do so.
Admittedly a small nitpick, that doesn't really affect the decision to merge.

chevin-ken: Agreed, just display the name

PR #141, opened 2019-11-11
Add waitlisting

jameslzhu:
This PR adds waitlisting by removing the requirement that rsvp count is under the event's `rsvp_count` cap, and allowing rsvps regardless.

Waitlist position is calculated dynamically, and is not stored anywhere. Waitlist position is 1-indexed (i.e. if nobody is in front of you but you are on the waitlist, you are in waitlist position 1). This is determined by counting the number of people before you (ordered by rsvp creation time), then slicing at the rsvp_count.

I've also removed some redundant authentication checking on the rsvp() view, which should be performed by the `@login_required` annotation anyways. 

Side note, should we post screenshots every time we make visual changes?

Screenshot from 2019-11-10 22-22-28

### DIFF
--- a/hknweb/events/templates/events/show_details.html
+++ b/hknweb/events/templates/events/show_details.html
@@ -74,10 +74,17 @@
         <p class="rsvp-list"> Count: {{ rsvps.count }} </p>
       {% endif %}
       {% for rsvp in rsvps %}
-        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user }})</p>
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.email }})</p>
       {% endfor %}
-    {% elif not rsvpd %}
+    {% else %}
       <p class="rsvp-list">No rsvps.</p>
+    {% endif %}
+
+    {% if waitlist %}
+      <h3 class="rsvp-list-title">Waitlist</h3>
+      {% for rsvp in waitlist %}
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.email }})</p>
+      {% endfor %}
     {% endif %}
   </div>
 </div>
--- a/hknweb/events/templates/events/show_details.html
+++ b/hknweb/events/templates/events/show_details.html
@@ -64,15 +64,14 @@
       <p class="rsvp-list">You have not rsvp'd</p>
     {% endif %}
 
-    
-    <h3 class="rsvp-list-title">RSVPs</h3>
-    
+
+    {% if limit %}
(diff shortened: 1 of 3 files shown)


hknweb/events/templates/events/show_details.html
```diff
@@ -64,21 +64,31 @@ <h1 id="event-detail-title"> {{ event.name }} </h1>
...
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.email }})</p>
       {% endfor %}
-    {% elif not rsvpd %}
+    {% else %}
       <p class="rsvp-list">No rsvps.</p>
     {% endif %}
+
+    {% if waitlist %}
+      {% if waitlisted %}
+        <h3 class="rsvp-list-title">Waitlist (your position: {{ waitlist_position }})</h3>
```
alexander-zw: The fact that you are waitlisted should be more obvious. We should not display "You are rsvp'd [checkmark]" since that is misleading, the person might think they are done. It should display something like "You are waitlisted [... image]" Making the text a special color won't hurt either.

hknweb/events/views.py
```diff
@@ -29,6 +28,7 @@ def is_cand_or_officer(user):
 
 def index(request):
     events = Event.objects.order_by('-start_time')
+    Event.objects.all()
```
alexander-zw: What is thing line doing?

hknweb/events/views.py
```diff
@@ -62,10 +71,8 @@ def rsvp(request, id):
     event = get_object_or_404(Event, pk=id)
     rsvps = event.rsvp_set.count()
 
-    if request.user.is_authenticated and (event.rsvp_limit is None or rsvps < event.rsvp_limit):
```
alexander-zw: To clarify, we didn't need to check user is authenticated in the first place because the decorator did that right?

hknweb/events/templates/events/show_details.html
```diff
@@ -64,21 +64,31 @@ <h1 id="event-detail-title"> {{ event.name }} </h1>
...
+
     {% if rsvps %}
-      {% if limit %}
-        <p class="rsvp-list"> Count: {{ rsvps.count }} / {{ limit }} </p>
-      {% else %}
-        <p class="rsvp-list"> Count: {{ rsvps.count }} </p>
-      {% endif %}
       {% for rsvp in rsvps %}
-        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user }})</p>
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.email }})</p>
```
alexander-zw: How does this look. I feel like the email is too long to display

alexander-zw:
Also there appears to be new alerts introduced

PR #146, opened 2019-11-15
Add colors to event types

### LINKED ISSUE #103: Color-code events by type
Perhaps add a specific RGB (#FFFFFF) color field to each event type?

jameslzhu:
Fixes #103.

~I think a legend is still not yet implemented.~

Legends implemented!

Screenshot from 2019-11-20 01-19-41

Also upgraded the fullcalendar version to v4.


alexander-zw:
We should make a legend

hknweb/events/admin.py
```diff
@@ -41,5 +44,3 @@ def cute_animal(self, request, queryset):
     cute_animal.short_description = "I wanna see a cute animal"
 
 admin.site.register(EventType)
```
alexander-zw: What if you add a color picker widget in the admin? I guess it's not really necessary though

jameslzhu: I haven’t found any widget to do so yet, but I have tinkered with the color picker in HTML5, which we do by setting an input attr to color in the form.

I’ll keep looking for this.

PR #147, opened 2019-11-15
Update Django to 2.2.7 LTS

jameslzhu:
I'd like to update Django to version 2.2, which is a Long Term Support version, expected to be supported to mid 2022. It also supports Python 3.8, which was released on Oct 14.

https://www.djangoproject.com/download/#supported-versions

I'm also updating mysqlclient, the MySQL connector library, to 1.4.5, cuz why not.

In some simple testing on my computer, nothing was broken, and we aren't using any features deprecated or removed in 2.2.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,8 +4,8 @@ verify_ssl = true
 name = 'pypi'
 
 [packages]
-django = "==2.1.11"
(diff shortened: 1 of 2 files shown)


jameslzhu:
Sad bears nobody reviews my pull requests

PR #149, opened 2019-11-20
Track interactivity requirements on candidate portal

alexander-zw:
Also some small CSS refactors etc


hknweb/candidate/migrations/0015_auto_20191119_2206.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.1.11 on 2019-11-20 06:06
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
alexander-zw: Uhh I don't remember making a migration not sure what this is

jameslzhu: I ran `makemigrations` on my computer and this re-appeared; it looks like you simply forgot to commit this migration in a much older change.

hknweb/candidate/views.py
```diff
@@ -48,6 +48,7 @@ def get_context_data(self):
         num_rejected = challenges \
                 .filter(Q(officer_confirmed=False) | Q(csec_confirmed=False)) \
                 .count()
+        num_confirmed = challenges.count() - num_pending - num_rejected
```
jameslzhu: Does this work? I've gotten errors about QuerySets being evaluated twice like this.

hknweb/settings/common.py
```diff
@@ -194,7 +194,17 @@
 
 # Constants for backend code
 
+# user groups
 CAND_GROUP = 'candidate'
 OFFICER_GROUP = 'officer'
 
+# default hard-coded event types for candidate semester
+# NOTE: these strings are also hard-coded in candidate/index.html
+MANDATORY_EVENT = 'mandatory'
```
jameslzhu: Was the change from `mandatory_meetings` to `mandatory` intentional?

hknweb/candidate/views.py
```diff
@@ -275,18 +281,28 @@ def sort_rsvps_into_events(rsvps):
...
+        settings.SERV_EVENT: 1,
+        settings.PRODEV_EVENT: 1,
+        settings.HANGOUT_EVENT: None,
     }
     req_statuses = dict.fromkeys(req_list.keys(), False)
     for req_type, minimum in req_list.items():
-        if len(sorted_rsvps[req_type])>= minimum:
+        num_confirmed = len(sorted_rsvps[req_type])
+        # officer hangouts are special case
+        print(req_type)
```
jameslzhu: Print statement?

hknweb/candidate/views.py
```diff
@@ -275,18 +281,28 @@ def sort_rsvps_into_events(rsvps):
...
+        if req_type == settings.HANGOUT_EVENT:
+            req_statuses[req_type] = check_interactivity_requirements(num_confirmed, challenges)
+            print(req_statuses)
```
jameslzhu: More print statement?

hknweb/events/templates/events/index.html
```diff
@@ -93,5 +93,6 @@ <h3  {% if messages.tags %} class="{{ message.tags }}" {% endif %} style="color:
         <button type='submit'>Add an event</button>
     </form>
     {% endif %}
+    <br>
```
jameslzhu: What does this <br> do? Just spacing at the bottom?

I'd personally be adjusting the padding / margin at the bottom of the content instead of adding a paragraph break here.

hknweb/candidate/views.py
alexander-zw: Uh hasn't errorred yet

hknweb/settings/common.py
alexander-zw: Yes, it made more sense to me because then it's consistent with what they call it on the user end, but I guess calling it mandatory meetings is more accurate

hknweb/candidate/views.py
alexander-zw: Oops

hknweb/candidate/views.py
alexander-zw: More oops

hknweb/events/templates/events/index.html
alexander-zw: Unfortunately we've already been doing this in a lot of other places... Do you feel strongly about this? Besides css sucks

PR #154, opened 2019-11-24
Google calendar integration

18praveenb:
Features:
1. Every time an event or RSVP is changed, clear and repopulate the Google calendar. Could become a performance bottleneck, not sure though.
2. The Google Calendar shows events, times, locations, descriptions, types, RSVPs, link to event detail page.

Setup instructions:
1. Create a google account w/calendar
2. Download the credential and edit the path in google_calendar.py to point to it.
3. It may do a browser popup for authentication. Not sure how this would work with the server. Maybe you could just do it on a local computer and copy the token to server? Anyway, the token will be created in token_path.
4. Edit the index.html for the calendar page to point to the new calendar. Note that this uses the **primary calendar** so don't use it with an account that already uses calendar for something else! (Or update the code ig) I have a calendar account dev.hkn.mu.calendar@gmail.com we could use.


18praveenb:
It would fail, because there are no credentials. Might want to do so gracefully though

bri25yu:
Not sure what the difference with https://github.com/compserv/hknweb/tree/googlecalendar is, not going to close as we definitely do want to work on this

bri25yu:
Closed by https://github.com/compserv/hknweb/pull/390

PR #155, opened 2019-11-27
Allow officers to edit events

alexander-zw:
Also small style changes to css and comments

More accurately those with edit permission can edit events, which for now is just the officer group


hknweb/events/views.py
```diff
@@ -105,3 +108,11 @@ def add_event(request):
             messages.success(request, 'Something went wrong oops')
             return render(request, 'events/add_event.html', {'form': EventForm(None)})
     return render(request, 'events/add_event.html', {'form': EventForm(None)})
+
+@method_decorator(permission_required('events.change_event', login_url='/accounts/login/'), name='dispatch')
```
jameslzhu: I think `login_url='/accounts/login/'` is actually the default:
https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-LOGIN_URL

hknweb/events/views.py
```diff
@@ -105,3 +108,11 @@ def add_event(request):
...
+class EventUpdateView(generic.edit.UpdateView):
+    model = Event
+    fields = ['name', 'slug', 'start_time', 'end_time', 'location', 'event_type',
+              'description', 'rsvp_limit']
+    template_name_suffix = '_update'
```
jameslzhu: Why is the template named `event_update` if it's at `/events/<id>/edit`?
(Is some kind of consistency worth it?)

hknweb/events/templates/events/event_update.html
```diff
@@ -0,0 +1,46 @@
...
+    <form method="post">
+        {% csrf_token %}
+        {{ form.as_p }}
+        <input type="submit" value="Update">
+        {% if messages %}
+            {% for message in messages %}
+                <br>
+                <span class="message">{{ message }}</span>
+            {% endfor %}
+        {% endif %}
```
jameslzhu: Should we be putting these messages into the `base.html`? These seem pretty common.

hknweb/events/templates/events/event_update.html
```diff
@@ -0,0 +1,46 @@
...
+        .content {
+            padding-top: 1.5em;
+            padding-bottom: 1.5em;
+        }
+    }
+    .centered {
+        max-width: 700px;
+        margin: 0 auto;
+    }
+</style>
```
jameslzhu: Should we be putting these into base.html or base.css if they modify elements from base.html?

hknweb/events/static/events/style.css
```diff
@@ -29,6 +29,10 @@
     vertical-align: middle;
 }
 
+.message {
+    color: rgba(166, 60, 68, 1);
```
jameslzhu: If we move messages to base.html, this should be moved to base.css as well.

hknweb/events/templates/events/show_details.html
```diff
@@ -32,6 +32,13 @@ <h1 id="event-detail-title"> {{ event.name }} </h1>
     <p><b>Location</b>: {{ event.location }}</p>
     <p><b>Date</b>: {{event.start_time|date:"D, N j, Y"}}</p>
     <p><b>Time</b>: {{event.start_time|time}} to {{event.end_time|time}}</p>
+
+    {% if can_edit %}
+      <p><a href="{% url 'events:edit' event.id %}">
+          Edit This Event
```
jameslzhu: Is title capitalization appropriate? `Edit this event` is consistent with the `Add an event` capitalization on /events anyways.

hknweb/events/views.py
alexander-zw: > Approved. I'd like to see the message code (which seems frequently copy+pasted) moved into base.css and base.html, and perhaps the .content css as well, but otherwise pretty sound.

The message might appear in different places though for different pages

alexander-zw: I guess we could change it to have a default place to put messages? Seems too complicated for this pull request though

hknweb/events/templates/events/event_update.html
alexander-zw: But we don't want it to apply for all pages right?

hknweb/events/templates/events/show_details.html
alexander-zw: Uhhh I like this capitalization style

hknweb/events/views.py
alexander-zw: > Why is the template named `event_update` if it's at `/events/<id>/edit`?
> (Is some kind of consistency worth it?)

I guess, the reason was that the django terminology is "update" but in hkn we tend to say "edit"

hknweb/events/views.py
alexander-zw: But being explicit isn't bad

hknweb/events/templates/events/show_details.html
alexander-zw: Besides "Go Back to Calendar" is capitalized

hknweb/events/views.py
alexander-zw: It's already inconsistent lol, add_event versus event_edit

PR #157, opened 2019-12-01
CSS Changes

solomonjoseph:
1. Redesigned login page to create more modern and elegant feel
2. Fixed alignments for sign up and hkn elections pages


Pipfile
```diff
@@ -4,14 +4,36 @@ verify_ssl = true
 name = 'pypi'
 
 [packages]
-django = "==2.2.7"
```
alexander-zw: Why did you change the pipfile?

hknweb/templates/registration/login.html
```diff
@@ -1,34 +1,148 @@
 {% extends "base.html" %}
 {% load static %}
+{% block header %}
+<style>
```
alexander-zw: I think it would be better to have these in a style.css file

Pipfile
solomonjoseph: I didn't mean to change the pipfile

hknweb/templates/registration/login.html
solomonjoseph: Resolved

Pipfile
GeoHutch: https://www.atlassian.com/git/tutorials/rewriting-history

Here's some info on rebasing bad changes that made it into a PR.

ochan1:
Old PR, and fork was deleted
Decision to Close the PR

PR #176, opened 2019-12-13
Fix RSVP twice bug

alexander-zw:


alexander-zw:
Props to @briana-jin-zhang

hknweb/candidate/models.py
```diff
@@ -49,10 +49,6 @@ def confirmed(self):
     def rejected(self):
         return self.officer_confirmed is False or self.csec_confirmed is False
 
-    @property
-    def rejected(self):
-        return self.officer_confirmed is False or self.csec_confirmed is False
-
```
alexander-zw: Oh oof this was duplicated? I never noticed

alexander-zw:
Moved to another pull request

PR #180, opened 2020-02-05
Rsvp twice fix

chevin-ken:
Fix rsvp twice bug


alexander-zw:
Never mind we'll fix the imports later

PR #181, opened 2020-02-10
Announcements for home page

jeffjohn33:


hknweb/models.py
```diff
@@ -28,3 +32,20 @@ def clean(self):
         if self.phone_number:
             self.phone_number = re.sub("[^0-9]", "",self.phone_number)
             self.phone_number = "("+self.phone_number[0:3]+") "+self.phone_number[3:6]+"-"+self.phone_number[6:]
+
+class Announcement(models.Model):
+    """
+    Model for an announcement. Created by VP or some other superuser.
+    Displayed on the candidate portal. The title will be displayed in bold,
```
alexander-zw: This docstring seems a bit inaccurate...

hknweb/templates/landing/home.html
```diff
@@ -60,6 +60,23 @@
     </div>
 
     <div class="tutoring-column">
+        <div class="tutoring-panel">
+            <div class="tutoring-title">
```
alexander-zw: Is there a reason this is being included in the tutoring section?

hknweb/models.py
jeffjohn33: Ok fixed comment

PR #185, opened 2020-02-12
cand forms due date

haolinzhu:


hknweb/candidate/admin.py
```diff
@@ -93,9 +93,15 @@ def set_visible(self, request, queryset):
 
     def set_invisible(self, request, queryset):
         queryset.update(visible=False)
-        
+
     set_invisible.short_description = "Set selected as invisible"
 
+class CandidateFormAdmin(admin.ModelAdmin):
+    fields = ['name', 'link', 'duedate']
```
alexander-zw: I really would prefer due_date but eh

hknweb/candidate/admin.py
```diff
@@ -93,9 +93,15 @@ def set_visible(self, request, queryset):
...
+    list_display = ['name', 'link', 'duedate']
+    list_filter = ['name', 'link', 'duedate']
+    search_fields = ['name', 'link', 'duedate']
```
alexander-zw: I don't think people usually search for dates, probably remove that

hknweb/candidate/admin.py
```diff
@@ -93,9 +93,15 @@ def set_visible(self, request, queryset):
...
     def set_invisible(self, request, queryset):
         queryset.update(visible=False)
-        
+
     set_invisible.short_description = "Set selected as invisible"
 
+class CandidateFormAdmin(admin.ModelAdmin):
+    fields = ['name', 'link', 'duedate']
+    list_display = ['name', 'link', 'duedate']
+    list_filter = ['name', 'link', 'duedate']
```
alexander-zw: Don't think there's a need to filter by link. Probably don't need to filter by name either since they should be unique most of the time.

hknweb/candidate/models.py
```diff
@@ -56,6 +56,15 @@ def rejected(self):
     def __str__(self):
         return self.name
 
+class CandidateForm(models.Model):
+
+    name           = models.CharField(max_length=MAX_STRLEN, default='')
+    duedate        = models.DateTimeField(default=timezone.now)
+    link           = models.CharField(max_length=MAX_STRLEN, default='')
```
alexander-zw: Maybe make this a text field instead? In case of really long links? I'm not sure on this one

hknweb/candidate/templates/candidate/index.html
```diff
@@ -54,13 +54,18 @@ <h2>Candidate Requirements</h2>
                 <li>Pay all dues (2 sets)</li>
                 <li>Complete all required forms</li>
                 <ul class="events-sublist">
-                    <li><a href="">Candidate Information Form</a></li>
+
+                    {% for form in candidate_form %}
+                        <li><a href={{ form.link }}>{{ form.name }}</a> (due {{ form.duedate }})</li>
+                    {% endfor %}
+
+                    <!-- <li><a href="">Candidate Information Form</a></li>
```
alexander-zw: Hmm commented out code...

hknweb/candidate/views.py
```diff
@@ -64,6 +64,9 @@ def get_context_data(self):
                 .filter(start_time__range=(today, today + timezone.timedelta(days=7))) \
                 .order_by('start_time')
 
+        candidate_form = CandidateForm.objects \
```
alexander-zw: This is a tiny nitpick but I think it should be `candidate_forms`

PR #186, opened 2020-02-14
Bitbyte activity support

alexander-zw:
You can create bitbyte activity confirmations on the admin, and candidates can view how many activities were confirmed.


Makefile
```diff
@@ -35,6 +35,10 @@ migrate:
 makemigrations:
 	HKNWEB_MODE='dev' pipenv run python ./manage.py makemigrations
 
+.PHONY: migrations
```
bri25yu: I like this move!

hknweb/candidate/admin.py
```diff
@@ -15,6 +15,7 @@ class OffChallengeAdmin(admin.ModelAdmin):
     list_display = ('name', 'requester', 'officer', 'officer_confirmed', 'csec_confirmed', 'request_date')
     list_filter = ['requester', 'officer', 'officer_confirmed', 'csec_confirmed', 'request_date']
     search_fields = ['requester__username', 'requester__first_name', 'requester__last_name', 'officer__username', 'officer__first_name', 'officer__last_name', 'name']
+    autocomplete_fields = ['requester', 'officer']
```
bri25yu: massive lmao

hknweb/candidate/admin.py
```diff
@@ -75,6 +64,23 @@ def csec_reject(self, request, queryset):
     csec_reject.short_description = "Mark selected as rejected (csec)"
 
 
+class BitByteActivityAdmin(admin.ModelAdmin):
+
```
bri25yu: It may make admin lives easier if we had an apt name for this (maybe some combination of the candidate name and the first few words of the activity description?)
hknalex bitbyte admin name_LI

For example, something like this is more helpful and accessible:
hknalex bitbyte admin name suggestion

Can maybe implement by creating `str` or `repr` functions for the [`BitByteActivity`](https://github.com/compserv/hknweb/pull/186/files?file-filters%5B%5D=.html&file-filters%5B%5D=.py#diff-f7910a17582b60486a0c61e7208c9eeeR60) class?

alexander-zw:
OOPS forgot to add logic that checks if you've done enough bit byte activities!

hknweb/candidate/admin.py
alexander-zw: Done

PR #189, opened 2020-02-27
Department Tours form

conhuang:


hknweb/tours/forms.py
```diff
@@ -0,0 +1,22 @@
...
+from .models import DepTour
+from django.contrib.auth.models import User
+from django.forms.widgets import SelectDateWidget
+# from django.contrib.admin.widgets import AdminDateWidget
+
+import datetime
+
+
+class TourRequest(forms.ModelForm):
+    date = forms.DateField(widget=SelectDateWidget(), label='Desired Date')
```
bri25yu: It's currently possible to request a dept tour before the current time. We should add some sort of check in so that there's less confusion on both sides of the dept tour request e.g. Rehan reads the time is for last month and not this month and he has to reach out to the dept tour requester to confirm that they meant this upcoming month. 

This applies to both the mm/dd/yyyy:
hknconnie dept tour pr date submitted time before current time
and to the hh/mm on the next line down:   
hknconnie dept tour pr date submitted time before current time today

hknweb/tours/forms.py
```diff
@@ -0,0 +1,22 @@
...
+    desired_time = forms.TimeField(help_text='hh:mm 24-hour time', label='Desired Time')
```
bri25yu: Since some students may be flying in from the East Coast or internationally, we should take into account time zones for scheduling a dept tour. Maybe specify that all of these times are in PST (-7:00 GMT) and/or put in some conversion like this https://www.timeanddate.com/worldclock/converter.html?iso=20200317T200000&p1=791. This particular website encodes listed time zones in the URL, so we can use that to our advantage!

hknweb/tours/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
+from . import views
+
+app_name = 'tours'
+urlpatterns = [
+    path('', views.tour, name='index'),
+    # path('confirm', views.confirm, name='confirmtour'),
```
bri25yu: Let's remove this line if it's not being used? If it's there as a placeholder, let's open an issue and mark it with a more detailed comment in the code or smth?

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
...
+    tour = DepTour.objects
+
+    context = {
+        'tour': tour,
+    }
+    return render(request, 'tours/index.html', context)
+
+def send_request_email(request, form):
+        subject = 'Department Tour Request'
+        officer_email = 'conhuang@hkn.mu'
```
bri25yu: Should we be sending this to deprel@hkn.mu?

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
...
+            tour = form.save(commit=False)
+            tour.save()
+            
+            # send_request_email(request, form)
+
+            messages.success(request, 'Your request has been sent!')
+            return redirect('/tours/')
+        else:
+            print(form.errors)
+            messages.error(request, 'Something went wrong! Your request did not send. Try again, or email deprel@hkn.eecs.berkeley.edu.')
```
bri25yu: Do we want to have more specific errors? For example, if the email they entered was invalid, the `TimeField` will error, but the error that's displayed doesn't include anything about the time syntax being incorrect. The user could then assume it's an error on our side, and email Deprel (which is borderline okay, but it's more work for Deprel). 

For example, if I enter this invalid time syntax: 
hknconnie dept tour pr time submitted invalid form
it only displays this on the top: 
hknconnie dept tour pr time submitted invalid form throws general error

hknweb/tours/models.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Model for a department tour
+    """
+
+    class Meta:
+        verbose_name = "Department Tour"
+
+    name            = models.CharField(max_length=MAX_STRLEN, default='')
+    email           = models.EmailField(max_length=MAX_STRLEN, default='')
+    # verify_email    = models.EmailField(max_length=MAX_STRLEN, default='')
```
bri25yu: Why did we comment out this code? It's not too particularly necessary, but it's definitely good to have, especially if in the future we want to send the requester a confirmation email or something. Probably ties in to [this](https://github.com/compserv/hknweb/pull/189/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.py&file-filters%5B%5D=No+extension#diff-c36ae4f1c5ec9bda34ddbd1926b73238R18) too, so I'll include them in the same comment.

hknweb/tours/models.py
```diff
@@ -0,0 +1,32 @@
+from django.db import models
+from django.utils import timezone
+from django.conf import settings
+# from django.contrib.admin.widgets import AdminDateWidget
```
bri25yu: Remove comments

hknweb/tours/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+MAX_STRLEN = 255 # default max length for char fields
+MAX_TXTLEN = 2000 # default max length for text fields
+
+#.save adds to database
```
bri25yu: Remove comments

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
...
+        msg.send()
+
+def tour(request):
+    form = TourRequest(request.POST or None)
+    if request.method == 'POST':
+        if form.is_valid():
+            tour = form.save(commit=False)
+            tour.save()
+            
+            # send_request_email(request, form)
```
bri25yu: Remove comments/old code!

hknweb/tours/forms.py
```diff
@@ -0,0 +1,22 @@
+from django import forms
+from django.conf import settings
+from .models import DepTour
+from django.contrib.auth.models import User
+from django.forms.widgets import SelectDateWidget
+# from django.contrib.admin.widgets import AdminDateWidget
```
bri25yu: Remove comment

hknweb/tours/views.py
bri25yu: Also, can we put deprel@hkn.mu instead of deprel@hkn.eecs.berkeley.edu? It's a really minor thing, but I just think deprel@hkn.eecs.berkeley.edu is kinda gross. On the other hand, `hkn.eecs.berkeley.edu` is obviously an email for HKN at Berkeley, so that may have more clarity that `hkn.mu` (since it's not common knowledge that our chapter is the Mu chapter).

PR #194, opened 2020-03-14
Finished waitlisting

chevin-ken:
Functionality that I have tested:

Accounts: 1, 2, 3
Event: limit of 1

Actions in Order:
1 rsvps (displayed on the rsvp list)
2 rsvps (correctly displays waitlist position of 1)
3 rsvps (correclty displays waitlist position of 2)
2 unrsvps (log back in to 3 and it correctly displays waitlist position of 1)
1 unrsvps (log back in to 3 and now it is no longer on the waitlist and is on the rsvp list)

Let me know if you find any other edge cases


hknweb/events/views.py
```diff
@@ -44,15 +44,30 @@ def index(request):
...
     event = get_object_or_404(Event, pk=id)
-
-    rsvpd = Rsvp.objects.filter(user=request.user, event=event).exists()
     rsvps = Rsvp.objects.filter(event=event)
-
+    rsvpd = Rsvp.objects.filter(user=request.user, event=event).exists()
+    waitlisted = False
+    waitlist_position = 0
+    if (rsvpd):
+        #Gets the rsvp object for the user
```
alexander-zw: I would prefer if we kept comment style consistent, with a space after the `#`

hknweb/events/views.py
```diff
@@ -68,11 +83,10 @@ def rsvp(request, id):
     event = get_object_or_404(Event, pk=id)
     rsvps = event.rsvp_set.count()
```
alexander-zw: LGTM.com flags this as unused variable

hknweb/events/templates/events/show_details.html
```diff
@@ -81,7 +83,7 @@ <h3 class="rsvp-list-title">RSVPs</h3>
         <p class="rsvp-list"> Count: {{ rsvps.count }} </p>
       {% endif %}
       {% for rsvp in rsvps %}
-        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user }})</p>
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.username }})</p>
```
alexander-zw: I think I already fixed this in the earlier pull request. Did you pull from master?

hknweb/events/templates/events/show_details.html
```diff
@@ -64,8 +64,10 @@ <h1 id="event-detail-title"> {{ event.name }} </h1>
         <input type="submit" value="RSVP" />
       </form>
     {% endif %}
-
-    {% if rsvpd %}
+    
+    {% if waitlisted %}
+      <p class="rsvp-list">You are waitlisted. You are at position {{waitlist_position}}</p>
```
alexander-zw: This feels a little wordy

hknweb/events/models.py
```diff
@@ -41,7 +41,13 @@ def __repr__(self):
 
     def __str__(self):
         return self.name
-
+    def admitted_set(self):
```
alexander-zw: Not enough padding around these functions

hknweb/events/models.py
```diff
@@ -41,7 +41,13 @@ def __repr__(self):
...
+        return self.rsvp_set.order_by("created_at")[:self.rsvp_limit]
+    
+    def waitlist_set(self):
```
alexander-zw: Also these two functions are not used anywhere. Maybe use them instead of duplicating the same code in views.py?

alexander-zw:
Also, should we display the waitlist and waitlist size? It would be useful to the hosts of the event, for one.

hknweb/events/templates/events/show_details.html
```diff
@@ -64,8 +64,10 @@ <h1 id="event-detail-title"> {{ event.name }} </h1>
...
+    {% elif rsvpd %}
```
alexander-zw: Also, should we display the waitlist and waitlist size? It would be useful to the hosts of the event, for one.

hknweb/events/templates/events/show_details.html
alexander-zw: Also, would we display the blue ... icon to be more visually outstanding?

hknweb/events/templates/events/show_details.html
```diff
@@ -81,7 +83,7 @@ <h3 class="rsvp-list-title">RSVPs</h3>
         <p class="rsvp-list"> Count: {{ rsvps.count }} </p>
       {% endif %}
       {% for rsvp in rsvps %}
-        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user }})</p>
+        <p class="rsvp-list">{{ rsvp.user.first_name }} {{ rsvp.user.last_name }} ({{ rsvp.user.username }})</p>
       {% endfor %}
     {% elif not rsvpd %}
       <p class="rsvp-list">No rsvps.</p>
```
alexander-zw: For consistency, maybe display the RSVP count and limit even when there are no RSVPs?

hknweb/events/templates/events/show_details.html
alexander-zw: Also, would we display the blue ... icon to be more visually outstanding?

hknweb/events/templates/events/show_details.html
chevin-ken: Haven't pulled, but this should be fine because it is the exact same change

PR #198, opened 2020-03-18
Serv/Outreach Static Pages

dinganthony:
- Made new outreach app, rerouted urls from landing page
- Created baseline serv static pages for each event, modeled off of the old website's infopages
- WIP: After bugging bridge for photos, add galleries


conhuang:
There is a small error with your url paths - if i try accessing the serv pages directly from the home page, it errors since the actual path is '/serv/eecsday':
Screen Shot 2020-03-25 at 3 53 08 PM
Screen Shot 2020-03-25 at 3 54 43 PM

conhuang:
This is minor but is it possible to keep the outreach tab open every time the user clicks on one of the links? Right now it defaults to opening the student services tab:
Screen Shot 2020-03-25 at 3 54 43 PM

but it would be super helpful to show this instead because right now it's slightly inconvenient to navigate between the different serv pages (have to click open the outreach tab first then click the link) and I imagine when people want to know about these, they're going to want to read them one after another:
Screen Shot 2020-03-25 at 3 55 06 PM

bri25yu:
Not sure how to solve the latter comment from @conhuang about keeping the outreach bar open. Resolved the first URL issue. Merging with compserv master and merging this PR in

PR #199, opened 2020-03-18
About HKN page

conhuang:


hknweb/static/css/style.css
```diff
@@ -0,0 +1,307 @@
...
+.table-icon {
+    width: 1.5em;
+    vertical-align: middle;
+}
+.what-about-subtitle {
+      font-size: 1.1em;
+      font-weight: 400;
+      opacity: 0.8;
+      line-height: 1.75em;
+    }
```
jvperrin: nit: Indentation is a bit wonky here

hknweb/templates/about/abouthkn.html
```diff
@@ -0,0 +1,90 @@
...
+{% endblock %}
+
+{% block content %}
+
+<div class='centered'>
+<body>
+  <div class='row'>
+
+    <div class='column left'>
+        <p class='what-about-subtitle'>Eta Kappa Nu (HKN) is the <a href='https://hkn.ieee.org/' target='_blank'>national Electrical and Computer Engineering honor society</a>. Founded in 1915, the Mu chapter at UC Berkeley is among the most active engineering societies at Berkeley. We strive to serve the Electrical Engineering and Computer Science (EECS) community at Berkeley by providing weekly tutoring hours, exam review sessions, department tours, course and staff evaluations, among many other services. Our two offices are located in 290 Cory and 345 Soda near the <a href='https://berkeley.edu/map' target='_blank'>northeastern corner of the UC Berkeley campus</a>.</p>
```
jvperrin: Wrapping long lines like this (and the other long lines in this file) would be nice as it would be easier to view changes and edit this file (and that shouldn't affect how the text looks when the page is rendered I believe). More of a nice-to-have though.

hknweb/templates/about/abouthkn.html
```diff
@@ -0,0 +1,90 @@
...
+  <!-- Tab 2 -->
+  <input type="radio" name="tabset" id="tab2" aria-controls="indrel">
+  <label for="tab2">Industrial Relations</label>
+  <!-- Tab 3 -->
+  <input type="radio" name="tabset" id="tab3" aria-controls="community">
+  <label for="tab3">Community</label>
+  
+  <div class="tab-panels">
+    <section id="services" class="tab-panel">
+      <div style='padding: 0% 5%;'>
```
jvperrin: Add this inline styling into the CSS file you've added in this review (especially as it's consistent across all the tabs from what I can see)?

hknweb/templates/about/abouthkn.html
```diff
@@ -0,0 +1,90 @@
...
+          <h2>Student Services</h2>
+          
+          <p><strong>Tutoring:</strong> Students who need advice or tutoring in EE and CS courses are encouraged to drop by for free assistance. Officers hold office hours from 11am to 5pm every weekday.</p>
+
+          <p><strong>Exam Preparation:</strong> HKN provides a database of past exams and review problems to prepare students for midterms and finals. In addition, look out for exam review sessions for lower-division courses.</p>
+          
+          <p><strong>Course Surveys and Peer Advising:</strong> HKN conducts course surveys for all classes offered by the EECS department every semester. The results for all professors and TAs who have opted-in are available online. Students who need more advice in choosing classes are encouraged to consult our officers during office hours.</p>
+      </div>
+      <div class='column half'>
+        <p align='right'><img id="maker-workshop" src="{% static 'img/maker workshop.jpg' %}" alt="HKN Maker Workshop" width=500></p>
```
jvperrin: Ideally I think you wouldn't have static widths on images like this, otherwise it might look strange or overflow on phones with narrow widths. I'm curious how this looks on an actual phone though as I haven't tested that.

hknweb/templates/base.html
```diff
@@ -151,7 +151,7 @@
       </label>
     </div>
     <div id="about-nav-bar" class="sub-nav-bar navi-container">
-      <a href="">Contact Us</a>
+      <a href="/about">About HKN</a>
```
jvperrin: Putting something like `{% url 'about' %}` or something here would be better in case it moves paths (and keeps all the url information in the `urls.py` file. I believe to do this though you'll need to add `name='about'` to the new path added in that file so this knows to find it by that name.

There's also a bunch of other stuff in this file that doesn't use the `url` template helper but probably should, but that's to do in another review instead.

conhuang:
Screen Shot 2020-03-23 at 8 29 36 PM
Screen Shot 2020-03-23 at 8 30 38 PM

Here is how one of the panels looks like. I'm not sure why the footer looks like that though, so some help for that would be great

hknweb/templates/about/abouthkn.html
```diff
@@ -0,0 +1,91 @@
...
+  <label for="tab2">Industrial Relations</label>
+  <!-- Tab 3 -->
+  <input type="radio" name="tabset" id="tab3" aria-controls="community">
+  <label for="tab3">Community</label>
+  
+  <div class="tab-panels">
+    <section id="services" class="tab-panel">
+      <div class='tabpadding'>
+        <div class='row'>
+        <div class='column half'>
```
jvperrin: Indentation is also a bit strange here (and cascades into some sections below)

Not a big deal though

PR #201, opened 2020-03-24
Department tours pt2

conhuang:
This is the department tours form. I made a new pull request because I messed up the commits in the last request. This one should contain only the department tour form.


conhuang:
Also, not really sure how to send emails (currently it's commented out of my code). This is the error that comes up when I try decrypting secrets.py.gpg
Screen Shot 2020-03-25 at 4 11 52 PM

hknweb/admin.py
```diff
@@ -6,6 +6,10 @@
 
 
 
+from .models import Announcement
+MAX_STRLEN = 85
+MAX_TXTLEN = 2000
```
jvperrin: Why are these needed here? I don't think they're used by anything in this file as far as I can tell

hknweb/models.py
```diff
@@ -9,6 +9,11 @@
 MAX_STRLEN = 85 # default max length for char fields
 MAX_TXTLEN = 2000 # default max length for text fields
 
+from django.utils import timezone
+
+MAX_STRLEN = 85
+MAX_TXTLEN = 2000
```
jvperrin: These are already defined just above to the same value, so I don't think these are needed here (same with the timezone import, that's also already done just before this)

hknweb/templates/base.html
```diff
@@ -101,7 +101,7 @@
       <span class="slash">/</span>
       <a href="">Course Guide</a>
       <span class="slash">/</span>
-      <a href="">Department Tours</a>
+      <a href="/tours">Department Tours</a>
```
jvperrin: You already have a name for these, so doing something like `{% url 'tours' 'index' %}` I believe should work here and make this more dynamic

hknweb/templates/base.html
```diff
@@ -194,7 +194,7 @@
         <span class="slash">/</span>
         Course Guide
         <span class="slash">/</span>
-        Department Tours
+        <a href="/tours">Department Tours</a>
```
jvperrin: (same with this as above)

hknweb/tours/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DepTour',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(default='', max_length=255)),
+                ('email', models.EmailField(default='', max_length=255)),
+                ('confirm_email', models.EmailField(default='', max_length=255)),
```
jvperrin: I don't think both of these need to be persisted in the database, after all their values should match to be able to submit the form so it's just duplicate information by the time it gets into the database itself, right?

hknweb/tours/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+                ('desired_time', models.TimeField(verbose_name='Desired Time')),
+                ('date', models.DateField(default=datetime.date.today, verbose_name='Desired Date')),
+                ('phone', models.CharField(default='', max_length=15, validators=[django.core.validators.RegexValidator(message="Enter a valid 10-digit phone number in the format:'+999999999'", regex='^\\+?1?\\d{10,15}$')])),
```
jvperrin: The example (`+999999999`) only looks like it has 9 digits? Ultimately I'd do something like `+1234567890` as an example so it's clearer how many digits there are at a glance, but that's not a big deal

hknweb/tours/models.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        verbose_name = "Department Tour"
+
+    name            = models.CharField(max_length=MAX_STRLEN, default='')
+    email           = models.EmailField(max_length=MAX_STRLEN, default='')
+    confirm_email   = models.EmailField(max_length=MAX_STRLEN, default='')
+    desired_time    = models.TimeField(verbose_name='Desired Time')
+    date            = models.DateField(default=datetime.date.today,verbose_name='Desired Date')
+    phone_regex     = RegexValidator(regex=r'^\+?1?\d{10,15}$', message="Enter a valid 10-digit phone number in the format:'+999999999'")
```
jvperrin: (same comment as above about the phone number)

It also looks like the regex supports a 10-15 digit phone number, which is totally fine, but it doesn't match the description which says only a 10 digit number

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
...
+    tour = DepTour.objects
+
+    context = {
+        'tour': tour,
+    }
+    return render(request, 'tours/index.html', context)
+
+def send_request_email(request, form):
+        subject = 'Department Tour Request'
+        officer_email = 'deprel@hkn.mu'
```
jvperrin: This works, but I'd probably go with the `hkn.eecs.berkeley.edu` domain here instead as it's just the more canonical name. Doesn't really matter for the functionality itself though. Same with the text below saying to email if there's a validation error

hknweb/views/landing.py
```diff
@@ -5,6 +5,8 @@
 from hknweb.events.models import Event
 # from hknweb.tutoring.models import Tutor
 
+from hknweb.models import Announcement
```
jvperrin: This appears to already be imported up above? (line 3)

Pipfile.lock
```diff
@@ -1,557 +0,0 @@
-{
```
jvperrin: I don't think you want to delete all this? At least there were no corresponding changes to the `Pipfile` that went with this to lead to a delete like this.

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
+from django.shortcuts import render, redirect, get_object_or_404
+from django.views import generic
+from django.core.mail import EmailMultiAlternatives
+from django.core.mail import EmailMessage
+from django.template.loader import render_to_string
+from django.views.generic.edit import FormView
+from .forms import *
```
jvperrin: lgtm has a good point here that this can pollute the local namespace with whatever is under the forms module, which can vary. Usually it's better to explictly specify here what functions/classes you actually need in this file and only import those

hknweb/tours/views.py
```diff
@@ -0,0 +1,54 @@
+from django.shortcuts import render, redirect, get_object_or_404
+from django.views import generic
+from django.core.mail import EmailMultiAlternatives
+from django.core.mail import EmailMessage
+from django.template.loader import render_to_string
+from django.views.generic.edit import FormView
```
jvperrin: As lgtm mentions, there's some unused imports here. Generally just good to not include those if you don't need them, but they aren't a problem for the functionality here, just can make things slower and make the code a bit harder to read/maintain.

Pipfile.lock
conhuang: forgot to run pipenv lock!

bri25yu:
Merged this with master to merge this PR in :)) Thanks for your work Connie!!

PR #214, opened 2020-04-15
Track mini fun events on candidate portal and display confirmed rsvp counts on candidate portal

chevin-ken:
minifun


chevin-ken:
The mini fun event type in the database must be defined as "Mini Fun"

bri25yu:
Incorporated into the logic of https://github.com/compserv/hknweb/pull/283

PR #226, opened 2020-04-26
Events Leaderboard

erictang000:
Events leaderboard that will show the all time ranking of users by number of events that have had their rsvp's confirmed. Edited views.py and added leaderboard.html, and added to the navbar under events (changes in base.html and urls.py).

### DIFF
--- a/hknweb/indrel/apps.py
+++ b/hknweb/indrel/apps.py
@@ -1,0 +1,5 @@
+from django.apps import AppConfig
+
+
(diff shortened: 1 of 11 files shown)


jameslzhu:
Why are entire files being marked as a diff when the changes are more restricted than that? Did some whitespace / EOL character get changed? This will probably make resolving the merge conflicts harder unless this is resolved.

PR #229, opened 2020-04-27
CSV Reader

krishnabhattaram:
Reads in a csv file with header being the group that the username should be added to. Adds all usernames under the header to that group, handling errors appropriately.


bri25yu:
Thanks for your work @krishnabhattaram :)) Resolved as per https://github.com/compserv/hknweb/pull/265

PR #230, opened 2020-05-01
A page with all the events where you can rsvp with one click

### LINKED ISSUE #136: Add a page with all of your rsvps
Best if you can have two columns, rsvp'd and not rsvp'd, and you can move events from one column to another with one click

alexander-zw:
closes #136

### DIFF
--- a/hknweb/events/forms.py
+++ b/hknweb/events/forms.py
@@ -1,5 +1,5 @@
 from django import forms
-from .models import Event, Rsvp
+from .models import Event
 
 
 class EventForm(forms.ModelForm):
--- a/hknweb/events/forms.py
+++ b/hknweb/events/forms.py
@@ -1,8 +1,5 @@
 from django import forms
-from hknweb.models import User
-from hknweb.models import Profile
-from .models import Event
-from .models import EventType
+from .models import Event, Rsvp
 
 
 class EventForm(forms.ModelForm):
--- a/hknweb/events/models.py
+++ b/hknweb/events/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.contrib.auth.models import User
 
+from hknweb.utils import get_semester
 
 class EventType(models.Model):
     type = models.CharField(max_length=255)
@@ -32,6 +33,13 @@ class Event(models.Model):
     created_by  = models.ForeignKey(User, on_delete=models.CASCADE, default=None)
     created_at  = models.DateTimeField(auto_now_add=True)
     # updated_at  = models.DateTimeField(auto_now=True)
+
+    @property
+    def semester(self):
+        """ A string representation of the candidate semester of this event.
+            Assumes that there are only spring and fall semesters, separated at 07/01.
+            Example: "Spring 2020" """
+        return get_semester(self.start_time)
 
     def get_absolute_url(self):
         return '/events/{}'.format(self.id)
--- a/hknweb/events/models.py
+++ b/hknweb/events/models.py
@@ -50,6 +50,14 @@ class Event(models.Model):
             return self.rsvp_set.none()
         return self.rsvp_set.order_by("created_at")[self.rsvp_limit:]
 
+    def on_waitlist(self, user):
+        if not self.rsvp_limit:
+            return False
+        return list(self.rsvp_set
+                        .order_by("created_at")
+                        .values_list('user', flat=True)) \
+                    .index(user.id) >= self.rsvp_limit
+
     def newly_off_waitlist_rsvps(self, old_admitted):
         """ old_admitted must be a set, not a QuerySet. QuerySets are mutable views into the database. """
         new_admitted = set(self.admitted_set())
--- a/hknweb/events/templates/events/all_rsvps.html
+++ b/hknweb/events/templates/events/all_rsvps.html
@@ -29,6 +29,8 @@
             <li class="rounded-text-box" style="background-color: {{ event_type.color }};">
                 {{ event_type.type }}</li>
         {% endfor %}
+        <img class="table-icon" src="{% static 'candidate/maybe.png'%}">
+        waitlist
     </div>
 
     <div class="aligned">
@@ -96,6 +98,9 @@
                                    style="background-color: {{ event.event_type.color }};">
                                     {{ event.name }}
                                 </a>
+                                {% if event.waitlisted %}
+                                    <img src="{% static 'candidate/maybe.png' %}" alt="waitlisted" class="table-icon">
+                                {% endif %}
                             </td>
                             <td align="center">{{ event.start_time }}</td>
                             <td align="center">{{ event.location }}</td>
--- a/hknweb/events/templates/events/all_rsvps.html
+++ b/hknweb/events/templates/events/all_rsvps.html
@@ -1,0 +1,106 @@
+{% extends "base.html" %}
+{% load static %}
+{% block title %} Your RSVPs {% endblock %}
+
+{% block header %}
+<link rel="stylesheet" href="{% static 'events/style.css' %}">
+{% endblock %}
+
+{% block heading %} Your RSVPs {% endblock %}
+
+{% block content %}
+<div class="parent">
+
+    <p class="no-vertical-margin">
+        See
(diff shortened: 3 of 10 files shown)


hknweb/utils.py
```diff
@@ -36,3 +37,20 @@ def get_rand_photo(width=400):
...
+def get_semester(date):
+    """ Returns a string representation of the semester of this timezone object.
+        Assumes that there are only spring and fall semesters, separated at 07/01.
+        Example: "spring 2020" """
+    season = "spring" if date.month < 7 else "fall"
+    return "{} {}".format(season, date.year)
+
+def get_semester_bounds(date):
+    """ Returns the two dates that bound the current semester.
+        Assumes that there are only spring and fall semesters, separated at 07/01. """
```
jameslzhu, line 52: Summer semester?

alexander-zw: We don't have candidates in the summer

jameslzhu: We do have summer course survey data, however.

I’d at least add a comment specifying this was intended for candidate events, and not for the other uses of the Semester model we were working on.

bri25yu: Also, maybe in the future we'd like to have custom/admin control over this date since it's a rigid `07/01`. Let's make an issue for this and for the point that James mentioned?

hknweb/utils.py
```diff
@@ -36,3 +37,20 @@ def get_rand_photo(width=400):
     with open(find("animal_photo_urls.txt")) as f:
         urls = f.readlines()
     return urls[randint(0, len(urls) - 1)].strip() + "?w=" + str(width)
+
+# date and time
+
+def get_semester(date):
```
bri25yu, line 43: Is this used anywhere? If not, let's remove it

hknweb/events/templates/events/all_rsvps.html
```diff
@@ -0,0 +1,124 @@
...
+                    <tbody>
+                    {% for event in rsvpd_events %}
+                        <tr>
+                            <td align="center">
+                                <a href="{% url 'events:detail' event.id %}" class="rounded-text-box"
+                                   style="background-color: {{ event.event_type.color }};">
+                                    {{ event.name }}
+                                </a>
+                                {% if event.waitlisted %}
+                                    <img src="{% static 'candidate/maybe.png'%}" alt="waitlisted" class="table-icon">
```
bri25yu, line 100: It's not immediately clear that this picture represents that the user is currently waitlisted for the event. Perhaps a legend or another signifier?

hknweb/utils.py
alexander-zw: I don't see why we need that, it's not like this date will change. If it does it won't change very often so keeping it hard coded should be fine

hknweb/events/views.py
```diff
@@ -23,6 +25,39 @@ def index(request):
...
+        rsvpd_event_ids = Rsvp.objects \
+                .filter(user__exact=self.request.user) \
+                .values_list('event', flat=True)
+        rsvpd_events = all_events \
+                .filter(pk__in=rsvpd_event_ids)
+        not_rsvpd_events = all_events \
+                .exclude(pk__in=rsvpd_event_ids)
+
+        for event in rsvpd_events:
+            event.waitlisted = event.on_waitlist(self.request.user) # Is this bad practice? idk
```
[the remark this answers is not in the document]
alexander-zw: Why is this necessarily a bad thing?

hknweb/utils.py
alexander-zw: I anticipated that it might be used soon, like if we add a semester field to events or support to filter by semester. Do we have to remove all unused utils?

hknweb/events/models.py
```diff
@@ -50,6 +51,14 @@ def waitlist_set(self):
             return self.rsvp_set.none()
         return self.rsvp_set.order_by("created_at")[self.rsvp_limit:]
 
+    def on_waitlist(self, user):
+        if not self.rsvp_limit:
+            return False
+        return list(self.rsvp_set
+                        .order_by("created_at")
+                        .values_list('user', flat=True)) \
+                    .index(user.id) >= self.rsvp_limit
```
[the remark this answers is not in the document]
alexander-zw: This is not part of the public UI, so I don't see why there should be a problem. The only time there would be an error is during development, and a ValueError seems informative enough in that case

hknweb/events/views.py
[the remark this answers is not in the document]
alexander-zw: The thing is, the waitlist is constantly changing when people RSVP, un-RSVP, or change the limit. Having multiple sources of truth is a gateway for bugs. We could make the only source of truth to make users stored as part of events instead of having RSVP objects but I don't think that's necessarily better

hknweb/events/forms.py
```diff
@@ -1,8 +1,5 @@
 from django import forms
-from hknweb.models import User
-from hknweb.models import Profile
-from .models import Event
-from .models import EventType
+from .models import Event, Rsvp
```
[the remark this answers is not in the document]
alexander-zw: Removed

hknweb/events/models.py
```diff
@@ -1,5 +1,6 @@
 from django.db import models
 from django.contrib.auth.models import User
+from django.utils import timezone
```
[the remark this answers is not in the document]
alexander-zw: Removed

hknweb/events/templates/events/all_rsvps.html
alexander-zw: Is this fine? 
Screen Shot 2020-07-19 at 6 21 21 PM

hknweb/utils.py
alexander-zw: Added comment

hknweb/events/models.py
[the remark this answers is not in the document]
bri25yu: I meant the 500 that's shown to the user. If they send us a user with an invalid user id, a 400 should be shown.

hknweb/utils.py
bri25yu: In both of those use cases, we would change the events database schema and add a `semester` field. Also, course surveys adds a `Semester` entity that can be utilized. 

In terms of anticipating use, we can remove it for now, and add it back when it is actually used (since it is already in Git). 

On the principle of keeping the codebase clean and as maintainable as possible, yes, we have to remove all unused utils.

hknweb/events/views.py
```diff
@@ -23,6 +25,39 @@ def index(request):
...
+        all_events = Event.objects \
+                .filter(start_time__gte=semester_start) \
+                .filter(start_time__lte=semester_end) \
+                .order_by('start_time')
+        if view_option == "upcoming":
+            all_events = all_events.filter(start_time__gte=timezone.now())
+        rsvpd_event_ids = Rsvp.objects \
+                .filter(user__exact=self.request.user) \
+                .values_list('event', flat=True)
+        rsvpd_events = all_events \
```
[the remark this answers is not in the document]
alexander-zw: Added

hknweb/events/models.py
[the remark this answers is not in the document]
alexander-zw: If who sends a user? This function is only called on users that are rsvp'd to an event, so there shouldn't ever be an invalid user ID

hknweb/utils.py
alexander-zw: I decided to just add the semester to the detail page.

hknweb/events/models.py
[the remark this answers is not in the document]
bri25yu: Honestly not too sure if this case would ever be hit, but imo it'd be better to have the error logic in place. Resolving since it's a minor thing.

hknweb/events/views.py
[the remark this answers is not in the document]
bri25yu: I think I misunderstood the code and I am wrong! Resolving.

hknweb/utils.py
bri25yu: LMAO ok solid, sounds good :))

bri25yu: image
Is it possible to capitalize `fall`?

PR #241, opened 2020-06-25
fixed the calendar stuff

Rishabhk2357:


Rishabhk2357:
I see what you're saying; I agree that the huge amount of static files is extremely awkward. However, if we're not using a build system like webpack, and we're using script tags in html pages, then according to the [v5 upgrade guide](https://fullcalendar.io/docs/upgrading-from-v4#pre-built-bundles), we have to use their prebuilt bundles. Thus, updating keeping everything at v4 would probably be best (the only problem would be if v4 eventually gets deprecated and stops being supported).

bri25yu:
Issue already resolved in a [different PR](https://github.com/compserv/hknweb/pull/242).

PR #242, opened 2020-07-09
Events calendar fix

bri25yu:
Specifying calendar version to 4.4.0 after fullcalendar.io upgraded to v5. This is a solid temporary solution and a solid semi-temporary solution (at least until fullcalendar deprecates v4). Worst case is that we may need to upgrade to the newest version in a couple of years.

### DIFF
--- a/hknweb/events/templates/events/index.html
+++ b/hknweb/events/templates/events/index.html
@@ -6,14 +6,14 @@
 {% block header %}
 <!-- Scripts which make the calendar work:-->
 <link rel="stylesheet" href="{% static "events/style.css" %}">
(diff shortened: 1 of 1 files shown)


Rishabhk2357:
> Specifying calendar version to 4.4.0 after fullcalendar.io upgraded to v5. This is a solid temporary solution and a solid semi-temporary solution (at least until fullcalendar deprecates v4). Worst case is that we may need to upgrade to the newest version in a couple of years.

For everyone else, the problem with switching is that we will either have to use a third-party css build software like web pack, or we'll have to install the fullcalendar bundle, which will be a hassle to manage with updates.

PR #244, opened 2020-07-12
Course surveys infrastructure

bri25yu:
[Course surveys infrastructure doc](https://docs.google.com/document/d/1UxesSlQWb7FkRjbqsmaaNZrHC4QZ9ip5aJwO8DvussI/edit#)

Initial PR for course surveys infrastructure. The purpose is to merge in the minimal necessary amount of code so that multiple people can be working on multiple facets of course surveys.


bri25yu:
> This pull request **introduces 3 alerts** when merging [aaa8514](https://github.com/compserv/hknweb/commit/aaa8514623da2047f12be2f49fb3519216239144) into [57d815b](https://github.com/compserv/hknweb/commit/57d815b738f7449732d830ea4d44f6304bccbfca) - [view on LGTM.com](https://lgtm.com/projects/g/compserv/hknweb/rev/pr-4515d3e49dc474da306ef2bc67cd2d6e8d218b41)
> 
> **new alerts:**
> 
> * 3 for Unused import

Removing as many unused imports as possible as per LGTM analysis. The unused imports in `academics/models.py` are required to register the models with Django

hknweb/academics/models/base_models.py
```diff
@@ -0,0 +1,5 @@
+from django.db import models
+
+class AcademicEntity(models.Model):
+    created_at = models.DateTimeField()
+    updated_at = models.DateTimeField()
```
jameslzhu, line 5: I'm not sold on the need for a base class just for time fields. Any features we might actually anticipate?

hknweb/academics/models/course_surveys.py
```diff
@@ -0,0 +1,29 @@
...
+    pass
+
+
+class Survey(AcademicEntity):
+    # reference attributes
+    survey_icsr = models.ForeignKey('ICSR', on_delete=models.CASCADE, related_name='survey_icsr')
+
+    # value attributes
+    num_students = models.IntegerField()
+    response_count = models.IntegerField()
```
jameslzhu, line 16: Change to [PositiveIntegerField](https://docs.djangoproject.com/en/3.0/ref/models/fields/#positiveintegerfield) to disallow negative students / responses?

hknweb/academics/models/course_surveys.py
```diff
@@ -0,0 +1,29 @@
...
+    is_private = models.BooleanField(default=True)
+
+
+class Rating(AcademicEntity):
+    # reference attributes
+    rating_question = models.ForeignKey('Question', on_delete=models.CASCADE, related_name='rating_question')
+    rating_survey = models.ForeignKey('Survey', on_delete=models.CASCADE, related_name='rating_survey')
```
jameslzhu, line 23:
```suggestion
    question = models.ForeignKey('Question', on_delete=models.CASCADE)
    survey = models.ForeignKey('Survey', on_delete=models.CASCADE)
```

Otherwise code which uses ratings may end up looking like `rating.rating_question` and `rating.rating_survey`.

hknweb/academics/models/icsr.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ICSR(AcademicEntity):
+    """
+    InstructorCourseSemesterRelation (ICSR)
+    """
+    # reference attributes
+    icsr_course = models.ForeignKey('Course', on_delete=models.CASCADE, related_name='icsr_course')
+    icsr_department = models.ForeignKey('Department', on_delete=models.CASCADE, related_name='icsr_department')
+    icsr_instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE, related_name='icsr_instructor')
+    icsr_semester = models.ForeignKey('Semester', on_delete=models.CASCADE, related_name='icsr_semester')
```
jameslzhu, line 14: Again, don't see the need for the `icsr_` prefixing.

hknweb/academics/models/logistics.py
```diff
@@ -0,0 +1,30 @@
...
+    SPRING = 'Sp'
+    SUMMER = 'Su'
+    FALL = 'Fa'
+    YEAR_SECTION_CHOICES = [
+        (SPRING, 'Spring'),
+        (SUMMER, 'Summer'),
+        (FALL, 'Fall'),
+    ]
+
+    year = models.TextField(max_length=100)
```
jameslzhu, line 29: A string year? Bold move. (Might work better for formatting but may cause surprise TypeErrors on comparisons to integer years.)

hknweb/academics/models/course_surveys.py
```diff
@@ -0,0 +1,29 @@
...
+from .base_models import AcademicEntity
+
+
+class Question(AcademicEntity):
+    pass
+
+
+class Survey(AcademicEntity):
+    # reference attributes
+    survey_icsr = models.ForeignKey('ICSR', on_delete=models.CASCADE, related_name='survey_icsr')
```
jameslzhu, line 12: See comment about redundant prefixing.

hknweb/academics/models.py
```diff
@@ -0,0 +1,3 @@
+from .models.course_surveys import Question, Rating, Survey
+from .models.icsr import ICSR
+from .models.logistics import Course, Department, Instructor, Semester
```
conhuang, line 3: Why do we need a models.py if there is a folder which contains all the models?

hknweb/academics/models/logistics.py
```diff
@@ -0,0 +1,30 @@
...
+    pass
+
+
+class Department(AcademicEntity):
+    name = models.TextField(max_length=200)
+    abbr = models.TextField(max_length=100)
+
+
+class Instructor(AcademicEntity):
+    instructor_id = models.TextField(max_length=500)
```
conhuang, line 16: Do we need to specify primary key for this?

hknweb/academics/models/base_models.py
bri25yu: Good question, I should have made a point of this in the commit description. Eventually this AcademicEntity class will contain entity-type-agnostic functions like `merge` and `separate`, along with several methods that those two functions will require e.g. `get_value_attributes` and `get_reference_attributes`, etc. There's not documentation about these functions as of this moment, but when there is, it can be found in the [Database operations design doc](https://docs.google.com/document/d/1Xa56DlWdIGqI_nUDE9MkZKB12zg2vpR9dpLgEFzzyVc/edit).

hknweb/academics/models/course_surveys.py
bri25yu: Both fields should be at least nonnegative. I can't see a scenario where `num_students` will ever be 0, so that will use a `PositiveIntegerField`. However, we have several cases where `response_count` is 0, so the plan is to provide validation so that `response_count` is only ever non-negative.

hknweb/academics/models/course_surveys.py
bri25yu: Unfortunately, if we use a `rating.question`, Django considers that a clash with the `Question` class (for example). This is my workaround, any suggestions to the naming?

hknweb/academics/models/icsr.py
bri25yu: See the previous comment

hknweb/academics/models/logistics.py
bri25yu: The formatting was exactly my consideration; I'll take a look and see what the hknweb precedent is. If it's integer years, then we'll switch over to integer years.

bri25yu: [`alumni/models.py/Alumnus.year`](https://github.com/compserv/hknweb/blob/57d815b738f7449732d830ea4d44f6304bccbfca/hknweb/alumni/models.py#L40-L43) uses an IntegerField with validation. [`tutoring/models.py/Slot.day`](https://github.com/compserv/hknweb/blob/57d815b738f7449732d830ea4d44f6304bccbfca/hknweb/tutoring/models.py#L43) also uses IntegerField. IntegerField with validation it is!

bri25yu: Leaving out any even slightly more complicated validation logic, leaving in a todo with a reference to the alumni example.

hknweb/academics/models/course_surveys.py
bri25yu: See above comment in response to comment about redundant prefixing.

hknweb/academics/models.py
bri25yu: My migrations originally errored out if I didn't import them explicitly here, but I just tried making the initial migration without the `models.py` file and everything worked out. My best guess is that I was having other errors before that could've caused Django to fall back on explicit definitions or maybe my filesystem structure was wrong, although not too sure. Removing the file now!

hknweb/academics/models/logistics.py
bri25yu: The default primary key is just a Django generated `id` field. If your question is "why isn't the `instructor_id` the primary key", it should be, whoops. There might be a bit of additional complexity down the road if we need to change primary keys from say a `first_name`/`last_name` hash to the instructors CalnetID, but this is warranted.

bri25yu:
> As a more important comment, all of these ForeignKeys have the default behavior ``on_delete=models.CASCADE`: if the pointed-to objects are deleted, so is the current object. This would mean that deleting a semester, for instance, will cause any associated ICSRs (and thus ratings) to be deleted.
> 
> To defend against this kind of a catastrophic accident, we may want a different policy, like [PROTECT](https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.PROTECT) which in this case forbids deletion of the semester unless all dependents are already deleted.
> 
> I know this has been the default in models but for something as important as course survey data, additional protection feels justified.

I see absolutely no downside to using PROTECT, and given some of its benefits which you listed, I think PROTECT is the clear better choice among the two. As for the choice between PROTECT and other `on_delete` possibilities, none of the others apply.

hknweb/academics/models/course_surveys.py
Rishabhk2357: I think that the slight amount of redundancy is ok; using a word other than "rating" for the prefix would just be confusing, and Brian mentioned the naming. Though I might be missing something, but I don't see why this would create a conflict.

bri25yu: Leaving in for now; we can always revisit it.

PR #246, opened 2020-07-23
Course surveys database rest api

bri25yu:
Jira Issue: [CSI-19](https://hkn-mu.atlassian.net/browse/CSI-19)
Design doc: [Database operations](https://docs.google.com/document/d/1Xa56DlWdIGqI_nUDE9MkZKB12zg2vpR9dpLgEFzzyVc/edit)

Remember to `make migrate` and `make update` before testing the changes!


hknweb/academics/serializers/course_surveys_serializers.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import serializers
+
+from ..models.course_surveys import Question, Rating, Survey
+
+
+class QuestionSerializer(serializers.ModelSerializer):
```
conhuang, line 6: Have you looked into `HyperlinkedModelSerializer`s? They use hyperlinks for foreign and primary keys instead of the key values, which would be convenient because we wouldn't need to explicitly specify the URLs in the frontend to these entities. What do you think?

hknweb/academics/serializers/course_surveys_serializers.py
```diff
@@ -0,0 +1,21 @@
...
+    class Meta:
+        model = Question
+        fields = ['id']
```
conhuang, line 9: This is a comment more for the Question model that i didn't consider in the last pull request - but should the question text be a field in the Question model instead of the Rating model?

hknweb/academics/serializers/logistics_serializers.py
```diff
@@ -0,0 +1,27 @@
+from rest_framework import serializers
+
+from ..models.logistics import Course, Department, Instructor, Semester
+
+
+class CourseSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Course
+        fields = ['id']
```
conhuang, line 9: Same kind of question as above, is there a reason you didn't include the name of the course in the Course model?

hknweb/academics/views/course_surveys_views.py
```diff
@@ -0,0 +1,25 @@
...
+class RatingViewSet(viewsets.ModelViewSet):
+    queryset = Rating.objects.all()
+    serializer_class = RatingSerializer
+    permission_classes = (HasPermissionOrReadOnly,)
+
+
+class SurveyViewSet(viewsets.ModelViewSet):
+    queryset = Survey.objects.all()
+    serializer_class = SurveySerializer
+    permission_classes = (HasPermissionOrReadOnly,)
```
conhuang, line 25: Could be helpful to add the get/post/put/delete skeleton methods in the ViewSets so it’s clear the functionality we want with these views?

hknweb/academics/serializers/course_surveys_serializers.py
bri25yu: There have been cases where question texts have changed throughout semesters, but still constitute the same "question essence". For example, this semester's question might be "How effective was this instructor?" and another semester's question might be "In terms of effectiveness, rate this instructor". While the question texts differ, they still count as the "same" question. So we still want to query by question entity (which is why we have an entity to begin with), but we want the question texts to be relevant only for a particular semester, so the `question_text` is on the Rating, but the Rating also references a FK Question. 

Why don't we just have the most recent question text? 
1) This is more on principle, but displaying the semester-appropriate question text is staying true to the biases and wordings of the time. Minute changes in wording, while constituting the same "question", can definitely change the resulting scores.
2) What if we want to go display or use previous question texts? It's hard for me to come up with a case that isn't incredibly contrived, but this also goes along with the idea of keeping all of the data we get. 
3) By putting the question text on Rating, we deliberately account for variations in question_text, rather than treating it as an edge case.

Why don't we have a list of question_texts for a particular question?
1) Logistics wise, question_texts are tied to a Rating ==> Survey ==> ICSR ==> Semester. If we had a list of question_texts, we'd have to associate each question_text with a particular Survey, ICSR, and semester, which is pretty much what a Rating already does, so we take advantage of the position that Rating is in. 

tl; dr: putting question_text on Ratings lets us be more fine-grained and adaptable, while also being simple-ish to implement.

hknweb/academics/serializers/logistics_serializers.py
bri25yu: Analogous response as above.

hknweb/academics/serializers/course_surveys_serializers.py
bri25yu: Done.

hknweb/academics/views/course_surveys_views.py
bri25yu: Added list of methods and reference documentation.

hknweb/academics/serializers/course_surveys_serializers.py
bri25yu: It just occurred to me that we can definitely have both a `question_text` attribute on `Rating` entities and a `current_question_text` for the most recent question text on `Question` entities. This would also be the pattern for `Course` objects, etc. Making an issue about that in Jira.

bri25yu: Jira issue [CSI-53](https://hkn-mu.atlassian.net/browse/CSI-53)

PR #252, opened 2020-08-25
Fix event edit datetime format bug and improve event edit form UI

alexander-zw:
Added datetime format before data population happens, changed form template a bit, extracted out a constant string into utils

### DIFF
--- a/hknweb/events/templates/events/event_edit.html
+++ b/hknweb/events/templates/events/event_edit.html
@@ -40,7 +40,19 @@
 <div class="centered">
     <form method="post">
         {% csrf_token %}
-        {{ form.as_p }}
+        {% for field in form %}
+            <div class="row">
+                {{ field.label }}
+                {% if field.field.required %}
+                    <span class="message">*</span>
+                {% endif %}
+                {{ field }}
+                <span class="message">
+                    {{ field.errors }}
+                </span>
+            </div>
+        {% endfor %}
+
         <input type="submit" value="Update">
         {% if messages %}
             {% for message in messages %}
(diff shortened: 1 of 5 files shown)


hknweb/events/views.py
```diff
@@ -151,6 +152,14 @@ class EventUpdateView(UpdateView):
     form_class = EventUpdateForm
     template_name_suffix = '_edit'
 
+    def get_initial(self):
```
Rishabhk2357, line 155: Where is this function used?

hknweb/settings/common.py
```diff
@@ -11,8 +11,10 @@
 """
 import os
 from django.conf.global_settings import DATETIME_INPUT_FORMATS
+from hknweb.utils import DATETIME_12_HOUR_FORMAT
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+
```
Rishabhk2357, line 17: Unnecessary line add (not crucial to change, but would be nice)

hknweb/events/templates/events/event_edit.html
```diff
@@ -40,7 +40,19 @@ <h1 id="add-event-title"> Edit Event: {{ event.name }} </h1>
 <div class="centered">
     <form method="post">
         {% csrf_token %}
-        {{ form.as_p }}
+        {% for field in form %}
```
Rishabhk2357, line 43: What was wrong with the formatting produce by "form.as_p"

hknweb/events/views.py
chevin-ken: I believe get_initial is something that Django uses to populate the form with pre-existing data before it is actually rendered on the template

alexander-zw: Yeah so it's used in the parent class

hknweb/settings/common.py
alexander-zw: Huh not sure where that came from

PR #255, opened 2020-10-03
changed calendars aspect ratio to 1.25, can fit 5 events in one cell now

sfChen1029:
This fixes the problem of events being hidden (assuming we have no more than 5 events in a day). Will also look into event ordering.


bri25yu (approved):
LGTM lmao

PR #256, opened 2020-10-03
Candidate COMPSERV-52 COMPSERV-53 COMPSERV-54 COMPSERV-77

bri25yu:
- Added `hknweb/candidate/utils.py` and `hknweb/candidate/constants.py` and refactored some logic into those files
 - Organized imports
 - Added color coding to candidate portal checklist event titles based off of their respective event-type color
 - Removed `hknweb/candidate/tests.py` since it wasn't being used
 - Added confirm button on officer challenge review portal
 - Refactored candidate portal index html to used for loop, enabled by refactoring view to html dataflow
 - Merged checklist and requirements section

hkn compserv candidate portal 01
hkn compserv candidate portal 02

PR #257, opened 2020-10-06
Update Candidate Portal requirements for Fall 2020

conhuang:
Changed a few requirements to reflect the new candidate requirements


bri25yu:
I think this has some dept tours stuff included in it? Also, is it possible that we merge in https://github.com/compserv/hknweb/pull/256 that reworks some of the candidate portal first D:

conhuang:
> I think this has some dept tours stuff included in it? Also, is it possible that we merge in #256 that reworks some of the candidate portal first D:

Hi! yes I was just looking through that pull request actually. Also all the commits were unexpected, I just got a new computer, and I am not sure how this all happened, trying to figure that out

PR #266, opened 2020-10-13
Events COMPSERV-45 COMPSERV-46 COMPSERV-48 COMPSERV-79 COMPSERV-80

bri25yu:
- Added functionality to allow creation of recurring events on calendar
- Added separation by types for All RSVPs page
- Added markdown capabilities to event descriptions
- Added clickable link functionality to event locations
- Added functionality for officers to confirm on the events detail page
- Added access level permissions functionality to events

compserv hknweb events details page changes
compserv hknweb events edit page changes
compserv hknweb events new page changes
compserv hknweb events all rsvps page changes


hknweb/events/templates/events/event_form_base.html
```diff
@@ -0,0 +1,81 @@
...
+        });
+        $(function () {
+            $("#id_end_time").datetimepicker({
+                format: 'm/d/Y g:i A',
+                formatTime: 'g:i A',
+                validateOnBlur: false,
+            });
+        });
+    </script>
+</div>
```
conhuang, line 81: This is amazing, thank you for making the code more concise

hknweb/events/models.py
```diff
@@ -22,8 +24,16 @@ class Event(models.Model):
     end_time    = models.DateTimeField()
     location    = models.CharField(max_length=255)
     event_type  = models.ForeignKey(EventType, models.CASCADE)
-    description = models.TextField()
+    description = MarkdownxField()
     rsvp_limit  = models.PositiveIntegerField(null=True, blank=True)
+    access_level = models.IntegerField(
```
Rishabhk2357, line 29: Is there a level for officers/should there be?

hknweb/events/templates/events/event_form_base.html
bri25yu: yeet!

hknweb/events/models.py
bri25yu: I originally thought there should be a distinction between officers and members, but then I was wondering how often (if at all) we would ever need that distinction (like are there any officer events that members shouldn't see?). What do you think?

bri25yu: Closing for now! (we'll probably end up doing access control in a less terrible way later on too lmao)

Rishabhk2357: Yeah this makes a lot of sense (sorry this slipped by).

PR #267, opened 2020-10-15
New candidate now has first and last name input

bri25yu:


bri25yu:
Pretty straightforward; just merging in automatically

PR #270, opened 2020-11-13
Raising z-index of location hover display to not be covered

bri25yu:


bri25yu:
Merging automatically since it's a one line fix

PR #271, opened 2020-11-13
Timezone display causing django display errors; removing for now

bri25yu:


bri25yu:
Merging automatically since it's a oneline fix

PR #272, opened 2020-11-13
Removing pipenv setup virtualenv

bri25yu:
Added a minimal development environment setup using `virtualenv`
- I didn't extensively test this, i just `make migrate` then `make dev` and checked that it worked, pretty much what our TravisCI does lmao (also i removed references in our travis.yaml file that broke our TravisCI whoops)
- Tried removing all traces of `pipenv`, but there are a few references in fabfile.py (like James was mentioning before) that I didn't want to touch -- need y'alls input on that


Makefile
```diff
@@ -6,53 +6,46 @@ PORT ?= 3000
 
 .PHONY: dev
 dev:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py runserver $(IP):$(PORT)
+	HKNWEB_MODE='dev' python ./manage.py runserver $(IP):$(PORT)
```
jvperrin, line 9: I'd suggest you still prefix this with the virtualenv location so that you don't get people accidentally using their system python if they haven't activated the virtualenv. Essentially like what was in f7ef64963aab581e53e8beb1b99c46d72c8b899d before #4, but don't worry about using venv-update.

requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+mysqlclient==1.4.5
+whitenoise==4.1.1
+pillow==6.2.0
+django-markdownx==2.0.27
+gunicorn==19.9.0
+fabric==2.4.0
+social-auth-app-django==3.1.0
+django-autocomplete-light==3.5.1
+
+pre-commit==1.14.4
```
jvperrin, line 11: I wouldn't mix the dev and prod requirements together. Often the dev ones are in a `requirements-dev.txt` instead.

It's also usually useful to have a place where they're pinned (like they are here) with all dependencies pinned too, and then have a place for the actual looser requirements directly needed for the app (django for instance, but not specifying a version unless there's compatibility problems), this makes it much easier to be able to upgrade and know why things are pinned.

.gitignore
```diff
@@ -109,3 +109,6 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+# VirtualEnv
+hknweb_dev
```
jameslzhu, line 114: usually I call this folder `venv` or `.venv`, `hknweb_dev` is pretty similar to the source folder `hknweb`.

Makefile
jameslzhu: Yeah, something like `<venv>/bin/python`.

.travis.yml
```diff
@@ -11,11 +11,4 @@ env:
...
   - HKNWEB_MODE=dev
 install:
-  # Dependencies
-  - pip install pipenv
-  - pipenv install
-  # Dev dependencies
-  - pipenv install --dev
 script:
-  - pipenv run python ./manage.py migrate
-  - pipenv run python ./manage.py check
```
jvperrin, line 21: I think you'll still need something here? Otherwise I don't think it'll know how to do any setup or run any tests

Makefile
```diff
@@ -6,53 +6,46 @@ PORT ?= 3000
...
-	@echo Please add $(PIP_HOME) to your PATH variable.
-
-.PHONY: venv
-venv: Pipfile Pipfile.lock
-	pipenv install --dev
+.PHONY: install
+install:
+	python -m pip install --upgrade pip
+	python -m pip install --upgrade setuptools
+	python -m pip install -r requirements.txt
```
jameslzhu, line 25: +1 prefix with `<venv_dir>/bin/python` instead of just `python`.

jameslzhu: Also consider installing `wheel,` it may fix some issues with binary packages.

Makefile
```diff
@@ -6,53 +6,46 @@ PORT ?= 3000
...
+	HKNWEB_MODE='dev' python ./manage.py shell
 
 .PHONY: test
 test: venv
-	pipenv run pytest -v tests/
-	pipenv run pre-commit run --all-files
-
-.PHONY: clean
-clean:
-	pipenv --rm
```
jameslzhu, line 51: Try just changing to `rm -rf ./<venv>`?

fabfile.py
```diff
@@ -92,6 +92,7 @@ def decrypt_secrets(c):
 def install_deps(c: Connection):
     print('-- Installing dependencies')
     with c.cd(c.release_path):
+        # TODO: Edit out pipenv dependency
         c.run('source .venv/bin/activate && pipenv install --deploy', echo=True, env={'PIPENV_VENV_IN_PROJECT': 'true'})
```
jvperrin, line 96: Here I think you just need to reach the pipenv command with `pip install -r requirements.txt`, and if the dev and prod requirements are split up then that'll do the same thing to only install the prod ones and skip any dev ones.

jameslzhu, line 96: Pretty sure this ought to be `.venv/bin/pip install -r requirements.txt`. The intent here is that if already in an activated virtualenv, pipenv will install into the shared virtualenv which `.venv` symlinks to `hknweb/prod/shared/venv`, so if you can re-leverage that it should still work.

README.md
```diff
@@ -43,18 +43,7 @@ which will attempt to safely shutdown the virtual machine, or kill it otherwise.
...
-
-On systems with sudo (root) privileges, you can run
-
-```sh
-$ pip install pipenv
-```
-
-For systems without sudo privileges, the following will work:
-
+Developming `hknweb` requires `venv`.
```
jameslzhu, line 46: Thank you kanye very informative

Pretty sure `venv` module is already included in standard Python 3.

Setup instructions here should go into detail about how to setup a virtualenv and how it works, which is kind of glossed-over currently. https://docs.python.org/3/library/venv.html

Vagrantfile
```diff
@@ -75,6 +75,7 @@ Vagrant.configure("2") do |config|
         python3 \
         python3-dev \
         python3-pip \
+        python3-venv \
```
jameslzhu, line 78: Not actually a requirement on Debian.

This package provides the `pyvenv` binary but you can still run `python3 -m venv` without it.

(Tested on my home desktop, which runs Debian 10 Buster).

Makefile
```diff
@@ -6,53 +6,46 @@ PORT ?= 3000
...
 
 .PHONY: livereload
 livereload:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py livereload $(IP):$(PORT)
+	HKNWEB_MODE='dev' python ./manage.py livereload $(IP):$(PORT)
 
-setup: pipenv venv migrate
+.PHONY: setup
+setup:
+	python3 -m venv hknweb_dev
```
jameslzhu, line 17: Try printing out instructions?

```make
@echo "When developing, activate the virtualenv with 'source hknweb_dev/bin/activate' so Python can access the installed dependencies."
```

jameslzhu: Also recommend using a Makefile variable for the actual virtualenv path.

```Makefile
VENV ?= hknweb_dev

...

$(VENV)/bin/activate ...

```

?= allows shell override of variable, but := works equally well.

.gitignore
bri25yu: `.venv` it is!

Makefile
bri25yu: Done

bri25yu:
Made a couple of edits, working more on this tomorrow

README.md
bri25yu: Added a bit more about why a virtual env is necessary, added link

Vagrantfile
bri25yu: Removed

Makefile
bri25yu: bless f7ef649, added prefixes

Makefile
bri25yu: Damn [pythonwheels](https://pythonwheels.com/) looks so cool. Not adding it in this time, adding in a comment with a TODO as a reminder.

.travis.yml
bri25yu: Not sure exactly what to put, I'll push something and see if it suffices lol

requirements.txt
bri25yu: Separated out prod and dev dependencies. Not sure I'm going to follow through with the pinned/unpinned version of our requirements at the moment, adding it as a TODO in the files.

fabfile.py
bri25yu: Resolving this convo to have everything in the [other duplicate thread](https://github.com/compserv/hknweb/pull/272/files?file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.txt&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&hide-deleted-files=true&file-filters%5B%5D=.lock#r522728333)

bri25yu: Replaced with the init logic similar to our `Makefile` and in `travis.yml`

.travis.yml
bri25yu: Looks like Travis passes!

Makefile
bri25yu: Added Makefile variables for the venv path and the echo instructions

Makefile
```diff
@@ -1,58 +1,61 @@
...
-	pipenv install --dev
+.PHONY: install
+install:
+	$(PYTHON) -m pip install --upgrade pip
+	# For issues with binary packages, consider https://pythonwheels.com/
+	$(PYTHON) -m pip install --upgrade setuptools
+	# TODO: pinned/unpinned dependency version.
+	# See https://github.com/compserv/hknweb/pull/272/files?file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.txt&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&hide-deleted-files=true&file-filters%5B%5D=.lock#r522718859
+	$(PYTHON) -m pip install -r requirements.txt
+	$(PYTHON) -m pip install -r requirements-dev.txt
```
jvperrin, line 31: I think you can combine the dev and prod requirements installs into one call by passing ,`-r requirements.txt -r requirements-dev.txt` (same with pip and setuptools if you'd like to)

Makefile
```diff
@@ -1,58 +1,61 @@
...
 .PHONY: shell
 shell:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py shell
+	HKNWEB_MODE='dev' $(PYTHON) ./manage.py shell
 
 .PHONY: test
 test: venv
-	pipenv run pytest -v tests/
-	pipenv run pre-commit run --all-files
+	pytest -v tests/
```
jvperrin, line 53: Should be able to do `$(BIN)/pytest` and `$(BIN)/pre-commit` here to use the virtualenv ones

Makefile
```diff
@@ -1,58 +1,61 @@
...
 .PHONY: dev
 dev:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py runserver $(IP):$(PORT)
+	HKNWEB_MODE='dev' $(PYTHON) ./manage.py runserver $(IP):$(PORT)
 
 .PHONY: livereload
 livereload:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py livereload $(IP):$(PORT)
-
-setup: pipenv venv migrate
```
jvperrin, line 15: I notice you've lost some make dependencies here, it used to do more setup (make a virtualenv, install packages, run migrations), where now it just created the virtualenv but doesn't install any packages). Maybe this should be a target called `venv` instead and `setup` could be the same as before?

fabfile.py
```diff
@@ -92,7 +92,12 @@ def decrypt_secrets(c):
 def install_deps(c: Connection):
     print('-- Installing dependencies')
     with c.cd(c.release_path):
-        c.run('source .venv/bin/activate && pipenv install --deploy', echo=True, env={'PIPENV_VENV_IN_PROJECT': 'true'})
+        # TODO: Edit out pipenv dependency
+        c.run("python3 -m venv .venv")
+        c.run("source .venv/bin/activate")
+        c.run(".venv/bin/python -m pip install --upgrade pip")
+        c.run(".venv/bin/python -m pip install --upgrade setuptools")
+        c.run(".venv/bin/python -m pip install -r requirements.txt")
```
jvperrin, line 100: I think deduplicating this with the Makefile would be good, but probably should not install the dev packages. That's a bit more difficult to do as-is and would probably need a new Makefile target that only installs the prod packages.

jvperrin (commented):
Seems pretty good to me, thanks for doing this and I've got just a few more comments but nothing too serious

bri25yu:
Add instructions to README on how to enable dev env e.g. `source .venv/bin/activate`

.travis.yml
```diff
@@ -11,11 +11,12 @@ env:
   # TODO: add test env (where DEBUG=False, secret key unencrypted)
   - HKNWEB_MODE=dev
 install:
-  # Dependencies
-  - pip install pipenv
-  - pipenv install
-  # Dev dependencies
-  - pipenv install --dev
+  - python3 -m venv .venv
```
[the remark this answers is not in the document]
bri25yu: Honestly did not know that Travis had access to Make, will try it out!

Makefile
```diff
@@ -1,58 +1,61 @@
...
+	HKNWEB_MODE='dev' $(PYTHON) ./manage.py livereload $(IP):$(PORT)
 
-.PHONY: pipenv
-pipenv:
-	pip3 install --user pipenv || pip install --user pipenv
-	@echo Please add $(PIP_HOME) to your PATH variable.
+.PHONY: setup
+setup:
+	python3 -m venv .venv
+	@echo "When developing, activate the virtualenv with 'source hknweb_dev/bin/activate' so Python can access the installed dependencies."
```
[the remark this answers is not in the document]
bri25yu: Fixed the echo info. Refactored all references of commands/directories into variables.

Makefile
```diff
@@ -1,58 +1,61 @@
...
-.PHONY: venv
-venv: Pipfile Pipfile.lock
-	pipenv install --dev
+.PHONY: install
+install:
+	$(PYTHON) -m pip install --upgrade pip
+	# For issues with binary packages, consider https://pythonwheels.com/
+	$(PYTHON) -m pip install --upgrade setuptools
+	# TODO: pinned/unpinned dependency version.
+	# See https://github.com/compserv/hknweb/pull/272/files?file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.txt&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&hide-deleted-files=true&file-filters%5B%5D=.lock#r522718859
```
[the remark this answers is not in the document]
bri25yu: Done

Makefile
bri25yu: Done

Makefile
bri25yu: Done

fabfile.py
bri25yu: Separated out the dev and prod dependencies into two separate commands, refactored fabfile to use the prod version

.travis.yml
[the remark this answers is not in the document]
bri25yu: Looks like it works!

Makefile
bri25yu: That's a good idea -- the current flow is `make setup` --> `source .venv/bin/activate` --> `make install`, but `make venv` makes a lot more sense

Makefile
```diff
@@ -1,58 +1,66 @@
...
 shell:
-	HKNWEB_MODE='dev' pipenv run python ./manage.py shell
+	$(MANAGE) shell
 
 .PHONY: test
 test: venv
-	pipenv run pytest -v tests/
-	pipenv run pre-commit run --all-files
+	$(BIN)/pytest -v tests/
+	$(BIN)/pre-commit run --all-files
```
jameslzhu, line 59: This may not have the intended effect, you may want to actually source the virtualenv here somehow. Strong suspicion this call will pick up the system Python.

Vagrantfile
jameslzhu: ... not actually removed?

fabfile.py
```diff
@@ -92,7 +92,10 @@ def decrypt_secrets(c):
 def install_deps(c: Connection):
     print('-- Installing dependencies')
     with c.cd(c.release_path):
-        c.run('source .venv/bin/activate && pipenv install --deploy', echo=True, env={'PIPENV_VENV_IN_PROJECT': 'true'})
+        # TODO: Edit out pipenv dependency
+        c.run("make venv")
+        c.run("source .venv/bin/activate")
```
jameslzhu, line 97: May want to double check this will actually activate the venv in the next call. IIRC Fabfile will actually separate each .run() call into a different SSH session, that's why I used && in the original.

That, or bake the venv sourcing into the makefile.

.gitignore
```diff
@@ -109,3 +109,6 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+# VirtualEnv
+.venv
```
jameslzhu, line 114: Need a trailing slash for a folder.

fabfile.py
bri25yu: Not sure how to bake the venv sourcing into the MakeFile -- I tried the naive just adding `source .venv/bin/activate` under `make venv`, but `make: source: Command not found`. Added all of them on the same line.

.gitignore
bri25yu: Not sure if we need to specify that `.venv` must be a directory, but adding it anyways for clarity.

Makefile
bri25yu: Welp, definitely is the case that there's an error `make: execvp: .venv/bin/pytest: Permission denied`. Tried `chmod +x $(BIN)/pytest`, but same error

fabfile.py
jvperrin: Ah, this'll be because commands in the Makefile are run in a separate shell with `/bin/sh`: https://stackoverflow.com/a/43566158

jameslzhu:
Wait you ... closed this without merging commits?

bri25yu:
Lmao was cleaning up stale branches and I guess I must've accidentally deleted this branch too

Makefile
bri25yu: Funnily enough, this is the same error that I got when trying to use Node module executables (not sure how I solved that issue either, but it's in the course surveys frontend branch [here](https://github.com/bri25yu/hknweb/blob/course-surveys-frontend/provision_frontend.sh)

jvperrin: I'm surprised, this works fine for me after creating a new virtualenv (`python3 -m venv .venv`) and installing pytest and pre-commit (`pip install pytest pre-commit`):
```
jvperrin@jaws:~$ .venv/bin/pytest --version
pytest 6.1.2
jvperrin@jaws:~$ .venv/bin/pre-commit --version
pre-commit 2.8.2
```

fabfile.py
jameslzhu: Hit 'em with the good old "bash -c "command".

Makefile
bri25yu: Yeah, even on a fresh install of the vm, this still occurs ://. Tried one more time `chmod -R 777 .venv` but still same permission denied error.

bri25yu: As this is the only issue, I'm going to just make an issue for this and merge.

PR #273, opened 2020-11-16
Fixing confirm RSVP access levels

bri25yu:
- Added event and user access levels (`event_access_level` and `user_access_level` respectively) to event details page
- Fixed corresponding uses in `show_details.html`


bri25yu:
Merging since it's a super small bug fix

PR #279, opened 2020-12-20
Removing login_url from common settings

bri25yu:


sammaher1 (approved):
Works. No longer causes error when logging out of settings menu.

PR #280, opened 2021-01-01
Markdown Security Patch

ochan1:


ochan1:
The original bug from this branch:
https://github.com/compserv/hknweb/pull/266

Used patch from https://github.com/neutronX/django-markdownx/issues/163

ochan1:
All "markdownify" function calls shall need to be called from the one made in "hknweb.utils" for security purposes

ochan1:
There is no need to modify the backend data, but if any malicious HTML code is pushed, it will stay on the backend

However, before showing up to the end-user, the text is sanitized

sammaher1 (approved):
Looks good, tried to inject various alerts and images into event descriptions and urls, which all did not succeed. Markdownify being added as a dependency seems like a worthwhile trade for the security and access to markdown.

PR #281, opened 2021-01-04
Vagrant Permissions to Owner for venv, Fabfile removed make venv in install_deps, and allow for command-line selectable target

ochan1:


ochan1:
To Test:
* Reclone another fresh copy of "hknweb" and with the new VagrantFile, try to Vagrant Up (preferably in both a Mac, Linux, and Windows machine)
* When there is another need to Deploy the website, try to deploy it with the new script and it should run normally. To test command-line selectable target, use: `fab deploy --target prod`, where "prod" is the only target available. It will AssertError out incorrect targets.

ochan1:
If you have a current Vagrant machine running, it is required that you follow these steps:
1. If you are pulling my code, pull to a different folder so it does not affect your current HKNWeb installation (Alternatively, if you want to run on your current hknweb code, you need to run `vagrant destroy` first (you will need to reinstall the Vagrant instance again))
2. If you have a ".venv" folder already, delete it (since you need to test if it can work on any person's first try in installing) [If you forgot, not a big deal. You can choose to skip this step if you want.]
3. Run `vagrant up` and let it finish
4. Follow all the instructions inside Setup in the Wiki
5. You should be able to now run Python Pip Packages like "fab" and "django-admin" inside the Virtual Environment

ochan1:
Just typing “fab” is okay multiple times (it just gives the Help stuff)

But only ONE person should try to deploy (I did it and it works, but a good idea for another person to try for experience and if I forget to push something) (But it may cause problems if multiple people deploy at the same time or a mass amount of the exact same deploy folder)

fabfile.py
```diff
@@ -176,13 +184,52 @@ def deploy(c, target=None, commit=None):
...
         update(c)
         publish(c)
         finish(c)
 
+# Please add the "target" parameter if you are adding more @task functions
+#  to allow custom targets to be used (regardless if your function itself will use it or not)
+
+
+# Source: https://realpython.com/python-command-line-arguments/#a-few-methods-for-parsing-python-command-line-arguments
+def parseArgs(systemArgs, argStart=1):
```
jameslzhu, line 199: Sir, this is a Wendy's

Can I interest you in [argparse](https://docs.python.org/3/library/argparse.html?highlight=argparse#module-argparse)? ([tutorial](https://docs.python.org/3/library/argparse.html?highlight=argparse#module-argparse)) I find it a lot more readable than custom sys.argv parsing. It's probably more robust to different argument ordering and even provides a help blurb.

fabfile.py
```diff
@@ -176,13 +184,52 @@ def deploy(c, target=None, commit=None):
...
+
+def getTarget(options, default=DEFAULT_TARGET):
+    target = options.get(TARGET_FLAG, default)
+    if target not in configs:
+        message = "\n\tTarget Configuration at {} is not a valid entry".format(TARGET_FLAG)
+        message += "\n\tInvalid Entry: " + target
+        assert target in configs, message
+    return target
+
+options = parseArgs(sys.argv)
```
jameslzhu, line 230: Note about variable style, I tend to follow [PEP 8](https://pep8.org/) which dictates snake_case for functions rather than camelCase. I'd prefer if we kept this consistent with the rest of the codebase.

In a perfect world we'd all `pip install black` and use [black](https://github.com/psf/black) to format all code but that may be asking a bit much. I'm not sure if black even works on Python 3.5.

fabfile.py
```diff
@@ -165,9 +169,13 @@ def publish(c: Connection):
 def finish(c):
     pass
 
+# For the following @task functions, "target" is an ignored parameter
+#  It is a workaround to allow for use in using command line arguments
+#  For selecting a "target" in ns.configure
+# Otherwise, fabfile will claim to not recognize it
```
jameslzhu, line 175: When I wrote this fab did not support arguments very well. This seems to have changed recently.
https://docs.fabfile.org/en/1.10/usage/fab.html#per-task-arguments

This may be a better alternative and let us avoid custom argv parsing entirely.

fabfile.py
```diff
@@ -176,13 +184,52 @@ def deploy(c, target=None, commit=None):
...
+        else:
+            opts[""].append(cmdArg)
+        i += 1
+    
+    return opts
+
+def getTarget(options, default=DEFAULT_TARGET):
+    target = options.get(TARGET_FLAG, default)
+    if target not in configs:
+        message = "\n\tTarget Configuration at {} is not a valid entry".format(TARGET_FLAG)
```
jameslzhu, line 225: '\t'? Really?

jameslzhu (commented):
A couple broad comments and some APIs you may want to look into.

fabfile.py
ochan1: I have an "assert target in configs, message" towards the end of the "if" statement
It's just for styling such that an error message comes out like this:

AssertionError ... :
     Target Configuration at ...
     Invalid Entry: NotATarget

ochan1: Rather than one long line of confusion

fabfile.py
ochan1: Not sure if that's okay with the ns.configure code to call inside each "task" function

The link describes a way to pass arguments into functions, so I probably still have to do arg parsing unless I can call like a "load_target" function and then a "run_deploy" function one after the other (The examples don't show that)

Also, I'm not too experienced with making updates (but please continue to link them so I can take a look in my spare time) for this CompServ system, so hopefully, a future iteration can maybe look into this or when I have time
Need to update, test, and repair SNAFUs if they happen

fabfile.py
ochan1: Styling Changed

fabfile.py
ochan1: Thanks
That was was easier than using "sys"

(Can I get fries with that though?)

haolinzhu (approved):
LGTM

PR #292, opened 2021-02-04
 Date range bug fix + Custom Title on Events

ochan1:


ochan1:
Related to a coursesemester.semester migration file
There was already a fix for it and rebasing is a bit weird since the "candidate" branch in ochan1 is very old
Will try on a different branch after merging into main ochan1 master

PR #302, opened 2021-03-05
Add CSV mass requirement checkoffs + exec group role

alvin-xu-5745:
- Added an exec group for Django admin (execs should still have officer role)
- Added a page (linked on user navbar) only accessible by officers that allows for mass event RSVP checkoffs with CSV
- Execs also have permissions to checkoff committee projects, dues, and forms.

image


hknweb/candidate/templates/candidate/checkoffs.html
```diff
@@ -0,0 +1,93 @@
...
+Mass Checkoffs
+{% endblock %}
+
+{% block content %}
+{% if viewer_is_an_officer %}
+<div class="parent">
+    <div class="table-div">
+        <h3>CSV Mass Checkoffs</h3>
+        {% if viewer_is_an_exec %}
+         - Here you can do mass checkoffs for an event, committee projects, dues, or forms. You have access to the latter three since you are an exec.<br>
```
ochan1, line 20: Small change, but we could maybe use "<ul>" and "<li>" as bullet points

hknweb/candidate/views.py
```diff
@@ -467,6 +467,109 @@ def add_cands(request):
...
+    users = []
+    for row in mem_csv:
+        try:
+            memberdto = CandidateDTO(row)
+        except AssertionError as e:
+            messages.error(request, "Invalid CSV format. Check that your columns are correctly labeled and filled out for each row.")
+            return redirect(next_page)
+        user = User.objects.filter(first_name=memberdto.first_name, last_name=memberdto.last_name, email=memberdto.email)
+        if not user:
+            messages.error(request, "Could not find user " + memberdto.first_name + " " + memberdto.last_name + " with email " + memberdto.email + ". Please check these parameters again.")
```
ochan1, line 543: Let's add a part here that says: No checkoff actions have been taken. Fix the errors and re-upload the entire file.

hknweb/candidate/views.py
```diff
@@ -467,6 +467,109 @@ def add_cands(request):
...
+        if checkoff_type == "event":
+            rsvp = Rsvp.objects.filter(event=event, user=user)
+            if rsvp.count() != 0:
+                rsvp = rsvp[0]
+                rsvp.confirmed = True
+            else:
+                rsvp = Rsvp.objects.create(user=user, event=event, confirmed=True)
+            rsvp.save()
+        elif checkoff_type == "project":
+            projectDoneEntry.users.add(user)
```
ochan1, line 558: For these 3 things, check if there is a duplicate
If there is, then don't add it (but you **don't** have to error out)

ochan1:
Hm, it looks like the CSV for Zoom combines the names together (or whatever is used on Zoom)
We could maybe consider a Tier 2 feature of helping people out with this, but if it's not always First and then Last we have a problem...

hknweb/candidate/views.py
ochan1: Never mind, Django won't add a duplicate for you

ochan1:
Besides cosmetic issues, everything else LGTM

Error checking works and adds people's RSVP and adds users checked off to the corresponding Checkoff Model

hknweb/candidate/views.py
```diff
@@ -467,6 +467,109 @@ def add_cands(request):
...
+
+def checkoff_csv(request):
+    if request.method != ATTR.POST:
+        raise Http404()
+    next_page = request.POST.get("next", '/')
+    csv_file = request.FILES.get("csv_file", None)
+    if not csv_file or not csv_file.name.endswith(ATTR.CSV_ENDING):
+        messages.error(request, "Please input a csv file!")
+        return redirect(next_page)
+    decoded_csv_file = csv_file.read().decode(ATTR.UTF8).splitlines()
```
ochan1, line 496: UTF-8 doesn't parse well for Mac
Can we use UTF-8-Sig?

ochan1:
LGTM, cosmetic problems resolved

PR #308, opened 2021-03-31
Bug Fix where Officer Hangout not tallying

ochan1:


ochan1:
Testing Instructions:
1.  Create two superusers, one acting as Officer and another as Candidate
2. Add an Event Type called "Hangout" (case sensitive)
3. Add an Event filed under "Hangout", with a date reasonable to the date range placed in Step 4 (if applicable). Do it inside Django Admin or Add Event, but you can set to "External" and allow you to skip Step 6
4. Add a Semester to use under CourseSemester, and set the Semester in the Setting
5. Add a "Requirement Hangout" and file under Officer Hangout. Leaving the date fields blank means an unlimited time range, which is independent to the Bug. I recommend a num_required to be high to be noticable.
6. Can be skipped if you set Event in Step 3 as External. Add a "Officer" Group and add your chosen officer accounts as "Officer"
7. Have the Candidate account RSVP to the Officer Hangout and the Officer account check it off
8. Check the portal and the "Remaining" should dock down

sammaher1:
Works well, counts down. Tested it using testing instructions.

PR #315, opened 2021-04-17
Shortlink Statistics

haolinzhu:
- Added modified_at and last_accessed_at fields to shortlink model
- Added more admin actions
- Added redirect to 404 page for invalid shortlinks


jdeng14:
created at, modified at, and last accessed at fields and 404 page all work as expected.

PR #317, opened 2021-04-26
completed leaderboard front-end feature

VincentWang001:
- Added leaderboard page in "events" section of website
- Added new url "leaderboard" path

### DIFF
--- a/hknweb/events/views.py
+++ b/hknweb/events/views.py
@@ -306,3 +306,22 @@ def send_off_waitlist_email(request, user, event):
                 settings.NO_REPLY_EMAIL, [user.email])
     msg.attach_alternative(html_content, "text/html")
     msg.send()
+
+
+# Leaderboard
+def show_leaderboard(request):
+    users = [
+        {'name': 'Abraham', 'points': 10},
+        {'name': 'Beta', 'points': 15},
+        {'name': 'Candy', 'points': 13},
+        {'name': 'Donuts', 'points': 20},
+        {'name': 'Person', 'points': 18},
+        {'name': 'Zebra', 'points': 100},
+        {'name': 'Penguin', 'points': -18},
+        {'name': 'Elephant', 'points': 20000},
+    ]
+    users.sort(key = lambda user: -user['points'])
+    context = {
+        'users': users
+    }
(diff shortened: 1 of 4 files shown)


hknweb/events/views.py
```diff
@@ -306,3 +306,22 @@ def send_off_waitlist_email(request, user, event):
...
+        {'name': 'Person', 'points': 18},
+        {'name': 'Zebra', 'points': 100},
+        {'name': 'Penguin', 'points': -18},
+        {'name': 'Elephant', 'points': 20000},
+    ]
+    users.sort(key = lambda user: -user['points'])
+    context = {
+        'users': users
+    }
+    return render(request, 'events/leaderboard.html', context)
```
ochan1, line 327: Please include a new line at the end of each file

hknweb/events/views.py
```diff
@@ -306,3 +306,22 @@ def send_off_waitlist_email(request, user, event):
                 settings.NO_REPLY_EMAIL, [user.email])
     msg.attach_alternative(html_content, "text/html")
     msg.send()
+
+
+# Leaderboard
+def show_leaderboard(request):
+    users = [
```
ochan1, line 313: TODO: Use real users and have a way to calculate points in the future

(Hint: Consider the "User" Django table)

hknweb/events/views.py
```diff
@@ -306,3 +306,22 @@ def send_off_waitlist_email(request, user, event):
...
+        {'name': 'Abraham', 'points': 10},
+        {'name': 'Beta', 'points': 15},
+        {'name': 'Candy', 'points': 13},
+        {'name': 'Donuts', 'points': 20},
+        {'name': 'Person', 'points': 18},
+        {'name': 'Zebra', 'points': 100},
+        {'name': 'Penguin', 'points': -18},
+        {'name': 'Elephant', 'points': 20000},
+    ]
+    users.sort(key = lambda user: -user['points'])
```
ochan1, line 323: Django has a way to sort, so when we filter and extract users, we can use that instead

bri25yu:
Thanks for your work @VincentWang001 :)) Resolved by https://github.com/compserv/hknweb/pull/365

PR #323, opened 2021-06-14
G cal link fix

sammaher1:
Fixed gcal link by urlifying the fields and updating the template link.

### DIFF
--- a/hknweb/events/constants.py
+++ b/hknweb/events/constants.py
@@ -6,12 +6,12 @@ DESCRIPTION_ATTRIBUTE_NAME = "description"
 LOCATION_ATTRIBUTE_NAME = "location"
 
 # See https://stackoverflow.com/questions/22757908/what-parameters-are-required-to-create-an-add-to-google-calendar-link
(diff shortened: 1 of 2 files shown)


anthonymaltsev:
I made a test event on a local server by logging into an admin account and creating an event through "Add An Event". Google calendar links seem to be fixed, which I think was the only thing changed in this PR. However, the new event that I made was not added to the events calendar on the website automatically (was only accessible by going to localhost:3000/events/1). I'm not sure if that is a problem with this PR because I've never made events before so I've never seen the intended behavior of adding an event.

sammaher1:
> I made a test event on a local server by logging into an admin account and creating an event through "Add An Event". Google calendar links seem to be fixed, which I think was the only thing changed in this PR. However, the new event that I made was not added to the events calendar on the website automatically (was only accessible by going to localhost:3000/events/1). I'm not sure if that is a problem with this PR because I've never made events before so I've never seen the intended behavior of adding an event.

Try setting the event access type to external. You may not be able to view if you set it to internal and are not logged in.

ochan1 (approved):

anthonymaltsev (approved):
LGTM

PR #333, opened 2021-10-04
Form failure keep entries, better form error format, better multiple day and location formatting

ochan1:


ochan1:
Pushing in for Candidate Semester immediate use

PR #334, opened 2021-10-05
Ability to Preprocess Candidates before saving them

ochan1:


ochan1:
Merging in for immediate use
The lgtm-com errors are normal since we have next_page with our request that is not user editable

PR #345, opened 2021-11-03
Candidate portal viewer

ochan1:
Ability to see another user's candidate portal and requirements checked off using: /cand/portal/<USERNAME>

Used visual and Ctrl+F to see where self.request and self.request.user is used and changed it so it is only for that one user

The Candidate Semester setting will be based off on whatever the user we are seeing their candidate semester set

To test, make multiple users, RSVP to some events each different to distinguish each, have each user see the Portal of the other

### DIFF
--- a/hknweb/candidate/admin.py
+++ b/hknweb/candidate/admin.py
@@ -21,7 +21,7 @@ from .models import (
     RequirementMandatory,
     RequirementMergeRequirement,
 )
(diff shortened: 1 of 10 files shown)


ochan1:
NOTE: Complete this Pull Request first to allow for rebasing of files: https://github.com/compserv/hknweb/pull/342

sidijju:
LGTM

Issue #311, opened 2021-04-02
Upgrading the HKNWeb Codebase to Python 3.7


ochan1:
Python 3.5 and the libraries associated with it is about to become depreciated

While there are newer versions of Python (like 3.9 at the time of this Issue writing), the OCF server as of 4/2/2021 only works with Python 3.5 globally and Python 3.7 via .venv

Currently, @ochan1 has made progress on getting Python 3.7 loaded to Vagrant and has gotten it to work, but...
* pip doesn't initialize properly, since it looks like Vagrant doesn't like it when pip writes a folder or something, causing a linking error for Python libraries

Reference my changes here: https://github.com/ochan1/hknweb/tree/python3_7_vagrant_upgrade

ochan1:
https://github.com/compserv/hknweb/pull/347

ochan1:
Clarification: the .venv on OCF uses Python 3.7 for some reason so this should work

ochan1:
Also, the fab file uses Python3.7 to make a venv at the start

https://github.com/compserv/hknweb/blob/master/fabfile.py#L145

ochan1:
Use Debian Buster: https://github.com/compserv/hknweb/pull/379

Discussions here: https://github.com/compserv/hknweb/pull/378

PR #347, opened 2021-11-09
Python 3.7 on VM, Pip upgrade workaround, Upgrade some dependencies

ochan1:
For the pip upgrade workaround, see: https://github.com/hashicorp/vagrant/issues/12057

The long story short is that a Virtual Machine File System, VBFS, doesn't delete files fully and causes an environment error, and states that a folder during upgrades of pip is not empty

This is a workaround for use on the Virtual Machine only and overwrites files during the upgrade rather than deleting them first

Also on top of having the VirtualMachine to use Python 3.7 due to the end of life of Python 3.5 + some dependency upgrades


bri25yu:
How should we go about testing that these changes don't break the website? e.g. re-deploying the website late at night and rolling back if something breaks

ochan1:
For the Python 3.7 and Pip workaround, it's only the Vagrant local system

I can see this being a problem for the Dependencies in the requirements.txt

We can revert to an older copy if the website breaks or push a manually reverted requirements.txt worst comes to worse

jameslzhu:
I would check what Python the OCF apphost is using and try to match their upgrade cadence, for dev/prod parity. Otherwise some newer Python features (f-strings, new methods, etc) might work in our VMs but break on prod.

jameslzhu:
Oh, looks like you've already checked this and are pushing a virtualenv 3.7 in #113.

Maybe look into pyenv to build and install a 3.7 Python environment on the apphost as a user install. I don't think .venv is intended for multiple Python versions.

ochan1:
OCF has Python 3.9, and Python 3.7 is only for Vagrant, so it's not too new compared to Python 3.9 so I'm assuming many dependencies should still support

ochan1:
FYI I deleted my latest comment
(Gonna delete this one as well, just wanted to utilize the email feature)

ochan1:
FYI I deleted my latest comment
Gonna delete this one as well

ochan1:
> Clarification: the .venv on OCF uses Python 3.7 for some reason so this should work

Added in #311

ochan1:
> Also, the fab file uses Python3.7 to make a venv at the start
> 
> https://github.com/compserv/hknweb/blob/master/fabfile.py#L145

Another one added in #311

PR #351, opened 2021-11-22
Candidate project (poll app)

NauqGnesh:


anthonymaltsev:
poked around, made some questions, answered them, and the results seemed to persist. clean code as well. nice job, looks good to me!

completely optional, but if you're interested in trying to do some more stuff on your own, you could try and integrate this more into the website (e.g. use the website HTML templates to make your polls look like the rest of the website; add some more user functionality, like a way to return to the main polls page after voting; adding links to polls from the main website; etc.)

PR #363, opened 2021-12-22
Adding indrel static pages

bri25yu:


bri25yu:
Thanks for your work @Matthew-Signorotti :)) Finally merging your static pages in!

PR #364, opened 2021-12-22
Event attendance leaderboard

bri25yu:
From https://github.com/compserv/hknweb/pull/226


bri25yu:
Wtf why are there so many differences?

PR #378, opened 2022-01-24
Back to Ubuntu Xenial + Python 3.7.3 built from Source in Vagrant

ochan1:
Ubuntu 16.04 via ubuntu/xenial64 reached End of Life (EOL) on April 30, 2021
As a result, Deadsnakes ended support for it recently (deadsnakes/issues#195)

The temporary fix was PR #377, but later discussions showed concern to try and remain as close to the OCF Ubuntu version as possible

So it was better to return back to Xenial, but with the desire to upgrade to Python 3.7, the alternate was to build from source, which is what this PR does (in addition to going back to Xenial)

Python 3.7.3 is used because that's the version OCF uses as of January 2022

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -103,8 +103,8 @@ Vagrant.configure("2") do |config|
     ln -s /usr/share/pyshared/lsb_release.py /usr/local/lib/python3.7/site-packages/lsb_release.py
 
     # Remove the Python executables (they done their purpose)
-    rm -rf Python-3.7.12
-    rm -rf Python-3.7.12.tgz
+    rm -rf Python-3.7.3
+    rm -rf Python-3.7.3.tgz
 
     # Back to main directory
     cd ~
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -101,11 +101,7 @@ Vagrant.configure("2") do |config|
     ln -fs /usr/local/bin/python3.7 /usr/bin/python
     ln -fs /usr/local/bin/python3.7 /usr/bin/python3
     ln -s /usr/share/pyshared/lsb_release.py /usr/local/lib/python3.7/site-packages/lsb_release.py
-
-    # Remove the Python executables (they done their purpose)
-    rm -rf Python-3.7.3
-    rm -rf Python-3.7.3.tgz
-
+    
     # Back to main directory
     cd ~
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -64,7 +64,6 @@ Vagrant.configure("2") do |config|
     ln -fs /usr/share/zoneinfo/US/Pacific /etc/localtime
     dpkg-reconfigure -f noninteractive tzdata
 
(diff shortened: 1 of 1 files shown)


ochan1:
For knowledge transfer, building more newer Python 3.7 versions does cause long test cases to be ran, which happens when the `./configure --enable_optimizations` is used

It is unknown how to skip those unless the `--enable_optimizations` flag is NOT used

jvperrin (commented):
Does this even switch back to xenial? I don't see a revert of #377 included in this

IMO using ubuntu bionic (18.04) and the python3.7 packages that ship with that distribution would be better for not needing to maintain this custom building in the future. It would also be close enough to the debian version the OCF uses (1 year later instead of xenial which is 1 year earlier)

Anyway, up to you, but that's what I would recommend anyway

Vagrantfile
```diff
@@ -7,11 +7,11 @@ Vagrant.configure("2") do |config|
   # Online documentation: https://docs.vagrantup.com.
   # Boxes: https://vagrantcloud.com/search.
 
-  # ubuntu/focal64 box used instead of debian/stretch64 because
```
ochan1, line 10: I forgot to update my master and rebase the branch to master, so the revert to xenial didn't show up
It shows up now

ochan1:
I considered using Bionic 18.07, but that bundles with Python 3.6 (https://dev.to/serhatteker/how-to-upgrade-to-python-3-7-on-ubuntu-18-04-18-10-5hab + https://packages.ubuntu.com/bionic/python3)

I also just saw the linked article in dev.to that shows a way to install Python 3.7 using apt-get without Deadsnakes (at least explicitly), but I also fear that the End of Life (EoL) problem will show up again and it's back to square one

ochan1:
I also forgot to rebase with the repo's master, so I did that just now

It now shows up in the Compare ([direct link to section](https://github.com/compserv/hknweb/pull/378/files/06f9024178f678d2494935c1e71ee1921ec18615#diff-1ea02434f746b6d1522cdef3d4a56e57cf2d5a72c5273927746a71eaf06d74fdR10-R14))

jameslzhu:
Is Debian Buster (10) an option? This Vagrant box contains the vboxsf kernel module for synced folders. https://app.vagrantup.com/debian/boxes/contrib-buster64

ochan1:
Sounds like a good option to address the Ubuntu choice (as mentioned in the comments) in the Vagrantfile

What you say also checks out in the Debian Vagrant link (https://app.vagrantup.com/debian) and when scrolled directly to the debian/contrib-buster64

OCF's upgrades looks also to be going underway as noted here: https://www.ocf.berkeley.edu/docs/staff/backend/buster/

I think Buster is an option! It's only two years older than what OCF uses and since it looks like OCF is gonna upgrade down the line too, I don't see why not

ochan1:
Decided to use Debian Buster https://github.com/compserv/hknweb/pull/379
Therefore closing PR and branch

PR #379, opened 2022-01-25
Use debian/contrib-buster64 + modifications - Python 3.7.3 support

ochan1:
Ubuntu 16.04 via ubuntu/xenial64 reached End of Life (EOL) on April 30, 2021
As a result, Deadsnakes ended support for it recently ([deadsnakes/issues#195](https://github.com/deadsnakes/issues/issues/195))

The temporary fix was PR #377, but later discussions showed concern to try and remain as close to the OCF Ubuntu version as possible

More discussion in #378, which lead to a suggestion by @jameslzhu to use Contrib Debian Buster in Vagrant (https://app.vagrantup.com/debian/boxes/contrib-buster64)
This has the Guest Additions feature needed for live updates (as mentioned in the Vagrantfile comments)

OCF's upgrades looks also to be going underway as noted here: https://www.ocf.berkeley.edu/docs/staff/backend/buster/

Buster is only two years older than what OCF uses and since it looks like OCF is gonna upgrade down the line too

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ install:
 	#  - https://github.com/compserv/hknweb/commit/124b108
 	#  - https://github.com/hashicorp/vagrant/issues/12057
 	# This is mainly a Vagrant workaround to allow pip upgrade
-	$(PYTHON) -m pip install --upgrade pip --ignore-installed
+	$(PYTHON) -m pip install --upgrade pip setuptools --ignore-installed
 
 	make install-prod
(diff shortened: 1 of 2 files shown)


ochan1:
Testing:
Basic testing with `vagrant up` and running the development environment startup commands (make migrate, make migrations, and make dev), and then playing the website for a bit

Also ran python3 which uses `python3.7.3` and also `which python` + `which python3` and both uses the `.venv` Python!

PR #389, opened 2022-02-24
Event forms check for end time

ochan1:
image

There was no mechanism to check if the end time is before the start time
Using the clean() function, as in the example in the [Django Documentation for form validation](https://docs.djangoproject.com/en/2.2/ref/forms/validation/), is appropriate here and the logic for this check is incorporated


bri25yu:
Are we going to merge this? lgtm

PR #414, opened 2022-03-07
fixing missing hangout attr

bri25yu:


hknweb/candidate/utils_candportal.py
```diff
@@ -488,12 +488,14 @@ def get_user_cand_data(self):
         context = {
             "announcements": announcements,
             "confirmed_events": {
-                event_key: confirmed_events[event_key]
-                for event_key in self.get_event_types_map(candidateSemester)
+                **{event_key: confirmed_events[event_key]
+                for event_key in self.get_event_types_map(candidateSemester)},
+                "hangout": confirmed_events["Hangout"],
```
ochan1, line 493: When the "Hangout" event tag is changed or deleted, this is gonna crash

bri25yu: It was gonna be wrong anyways due to https://github.com/compserv/hknweb/blob/29650ef34a10a7b1308fbc0f1274c89b8969f040/hknweb/candidate/utils_candportal.py#L295

ochan1: The code makes sure an EventType for "Hangout" exists first, then have the code (via the required_events dictionary) start to track the "Hangout" events

bri25yu: Right, so it doesn’t crash but is still wrong if we rename the “Hangout” event type. I’m ok with it crashing with the new change if we change the name to something else so at least it doesn’t fail silently.

ochan1: It's also a minor developer annoyance, I remember having to always add an Event or EventType when I was working on something related to the Candidate Portal before the flexibility feature was implemented

ochan1: https://github.com/compserv/hknweb/pull/416

PR #430, opened 2022-03-13
For course survey ratings: use a continuous/interpolated color spectrum instead of three discrete color options

Gabe-Mitnick:
Before (example with discrete colors)
image

After (dummy course with continuous colors)
image


bri25yu:
lgtm!

PR #431, opened 2022-03-14
MacBook Dependency Installs in Makefile - for M1 mainly

ochan1:
Reference link: https://betterprogramming.pub/managing-virtual-machines-under-vagrant-on-a-mac-m1-aebc650bc12c

As noted in the reference link, there is no current method to have Vagrant run on ARM computers
As of now, ARM computers only exists as the recently new M1 MacBooks

Since the only current source in the mass public is M1 MacBooks, this solution allows a Makefile guided install on the dependencies needed for MacBooks
The commands here are written for now as a start and there is an expectation to grow the list (especially the MacBook dependencies installed through "homebrew")

Linux users can already follow the Vagrantfile for installs

Alternates include Docker or an x86 Linux emulation as listed in the Reference Link, and so far these are easier methods to develop on for now and according to my ability

There are of course alternates to get this running on ARM, but this commit's author has no such resources to research further

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ VENV := .venv
 BIN := $(VENV)/bin
 PYTHON := $(BIN)/python
 MANAGE := HKNWEB_MODE='dev' $(PYTHON) ./manage.py
+BREW_DIR := $(shell brew --prefix)
 
 IP ?= 127.0.0.1
 PORT ?= 3000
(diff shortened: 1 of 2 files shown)


ochan1:
Replaced with this PR: https://github.com/compserv/hknweb/pull/491

PR #441, opened 2022-03-30
Add Media symlinking line in FabFile

ochan1:
Related to #435 

Moved Media folder to a private folder so it won't get accessed easily via the OCF public_html

Side note: Need to determine how to present Media that is public facing
A recommendation is to use: https://github.com/edoburu/django-private-storage (but how about download when needed?)


ochan1:
Tested with a deployment done on 3/30/2022 at 1:30 AM
Media folder is symlinked on the new website deployment

PR #446, opened 2022-04-05
Event Repeat Renaming + Relogic, Event Repeat Test, Default None for Empty Event Types in All RSVP

ochan1:
Event Repeat Renaming:
image

In this PR:
* Event Repeat Renaming + Relogic
* Event Repeat Test
* Default None for Empty Event Types in All RSVP (mainly for the Tabular Event test)

(Does not fix test failures from incorrect Response Code)

### DIFF
--- a/hknweb/events/tests/views/event_transactions/test_repeat_event.py
+++ b/hknweb/events/tests/views/event_transactions/test_repeat_event.py
@@ -1,0 +1,87 @@
+from datetime import datetime, timedelta
+from django.test import TestCase
+
+from django.urls import reverse
+
+from hknweb.events.tests.views.event_transactions.utils import setUp
+from hknweb.events.utils import generate_recurrence_times
+
+class RepeatIntervalsUtilsTests(TestCase):
+    def test_repeat_time_intervals(self):
+        start_time = datetime(2022, 3, 15, 11, 30)
+        end_time = datetime(2022, 3, 15, 3, 30)
+
+        def base_tests():
+            num_repeats = 0
+            period = 0
+            answer = [(datetime(2022, 3, 15, 11, 30), datetime(2022, 3, 15, 3, 30))]
+            result = generate_recurrence_times(start_time, end_time, num_repeats, period)
+            self.assertListEqual(result, answer, msg="No repeat and No period")
+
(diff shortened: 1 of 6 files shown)


hknweb/events/tests/views/event_transactions/test_repeat_event.py
```diff
@@ -0,0 +1,87 @@
...
+
+from hknweb.events.tests.views.event_transactions.utils import setUp
+from hknweb.events.utils import generate_recurrence_times
+
+class RepeatIntervalsUtilsTests(TestCase):
+    def test_repeat_time_intervals(self):
+        start_time = datetime(2022, 3, 15, 11, 30)
+        end_time = datetime(2022, 3, 15, 3, 30)
+
+        def base_tests():
```
bri25yu, line 14: What's the reasoning behind wrapping this in base_tests?

bri25yu:
Also do we want to keep the event recurrence code? Not sure if anyone actually uses it (looking at our calendar).

hknweb/events/tests/views/event_transactions/test_repeat_event.py
ochan1: Nothing special, just grouping them as a name of a series of small base case sanity checks

bri25yu: Why not just make them a test case or unwrap them?

ochan1: Agreed on making it a separate one, just did it

ochan1:
I agree it will be available rarely, but I think some uses cases include the LEG DAY event the towards the end of the Spring 2022 semester + Brownbag Talks in the Fall 2021 semester

Case example too is that I don't use recurring events often on Google Calendar, but it is super helpful when I make weekly events or something for meetings (especially CompServ meetings during my Fall 2021 and Spring 2022 semester)

My opinion is to keep it there and people can use it or skip it, though I do agree it's usage is rare

bri25yu:
Also, if we're gonna be putting the committee meetings on the calendar, we actually have a use case for the feature, I'm down to merge this in

PR #450, opened 2022-04-11
Checkpoint

aezhang:
I finished the Django Tutorial, Please let me know if there are any changes that need to be made.


bri25yu (changes requested):
Hi Andrew! Your project looks pretty good overall :)) A few things -- 
* Can you make the polls app a part of the hknweb project? So like you shouldn't need a separate `manage.py`, `settings.py`, etc and you should also need to modify some of the hkn repo files
* Make sure to run `make format` to format your code!

bri25yu:
Looks great to me! Cheers on finishing your candidate semester Andrew :))

PR #452, opened 2022-04-11
[Checkpoint] Tutorial part 1

agoldstein03:
This is to satisfy the HKN candidate checkpoint requirement.


hknweb/polls/views.py
```diff
@@ -0,0 +1,4 @@
+from django.http import HttpResponse
+
+def index(request):
+    return HttpResponse("Hello, world. You're at the polls index.")
```
ochan1, line 4: Please add a new line for every file
Looks like you did and forgot this one

ochan1 (changes requested):
Good progress on Part 1
Please continue as much as you can toward Part 7

hknweb/polls/views.py
agoldstein03: Fixed for next PR

PR #457, opened 2022-04-25
Work to make Course Guide better

ochan1:
Fixes to make Course Guide look better
Completed alongside work for Oscar's Info 247


ochan1:
Sample image
image

PR #458, opened 2022-04-26
Create polls app

uvikas:
Created polls app following django tutorial


anthonymaltsev:
Good job, everything looks good to me.

If you -- completely optionally -- want to take this project further you might find it interesting to implement the front end to make the polls accessible from the main website rather than just by URL and to make the page look like the rest of the website.

PR #460, opened 2022-04-27
[Do Not Merge] Polls App Candidate Project

agoldstein03:
Completed parts 1-7 of the polls project.


jyxzhang:
Looks good to me!
Very pertinent GIF on the polls page
Don't see any issues with style
An optional additional step would be to make the polls page look like the rest of the site and also make the polls page accessible from the main page (once again, optional). What you have now fulfills the candidate project requirement.
Hope to see you around in HKN!

PR #464, opened 2022-05-05
Candidate revamp

bri25yu:


ochan1:
Bug 1: Long "Proof" description
image

Bug 2: Long "Proof" link
image

ochan1:
Also, by removing the "Mass Checkoff" feature, it is reasonable with the password feature that events are easier to check off, even though passwords could be distributed, but what about for Forms and Due Payments? Candidates wanted the ability to see if their forms and payments were confirmed by HKN.

I remember we were talking about making an account for all people invited, which leads to how are we gonna do that? I don't see how a feature similar to Mass Creation of Candidates would be designed (see hknweb/candidate/views/mass_add_cands.py). Also, I remember why I wanted to push back a bit on the idea of making an account for all people invited because we can't just use people's data to create accounts without their permission and once they found out, they can be like "oh I didn't want this account created" or "I don't want to initiate HKN" and whatnot (a bit of data ethics precedence and it may scratch a bit on us using other people info without their permission with laws probably on that stuff based on precedence).
My recommendation is to keep this feature and have students fill out a Google Form or something on the website to request the creation of the data OR have a preapproval process first (and I'm not talking about going to Admin and checking off everyone who is a Candidate ... that was annoying back in Spring 2021 during my officer time)

Let alone I don't think such a reasonable, secure, and more legal replacement feature exists yet in your PR, so I'm pushing back against its deletion for now

-----

Also, the VP or CSec confirm of challenges thing is there for a reason from the past (maybe to make sure the challenge meets the requirement of not being from a committee officer or from Byte or smth), but I'd consult VP and CSec if they are okay with this feature removed ... I'm neutral about this change as long as it goes through them

bri25yu:
> Bug 1: Long "Proof" description image
> 
> Bug 2: Long "Proof" link image

Oversight on the original model design -- we have the proof max length set to 2000 characters. Will set to 100 characters now (definitely enough for a link).

bri25yu:
> By removing the "Mass Checkoff" feature, it is reasonable with the password feature that events are easier to check off, even though passwords could be distributed, but what about for Forms and Due Payments? Candidates wanted the ability to see if their forms and payments were confirmed by HKN. Would be annoying for VP or Tres to check them off individually.
The pipeline for this is I'm gonna do it first for the events RSVP confirm panel, changing the status and action columns into a single checkbox column. On click it'll do the correct confirm/unconfirm action. Then, we'll use that same template code to check people off for the other reqs. However, a precursor to all of this might be to consolidate all the html in hknweb and I am not ready for it LMAOOO
> 
> I remember we were talking about making an account for all people invited, which leads to how are we gonna do that? I don't see how a feature similar to Mass Creation of Candidates would be designed (see hknweb/candidate/views/mass_add_cands.py). Also, I remember why I wanted to push back a bit on the idea of making an account for all people invited because we can't just use people's data to create accounts without their permission and once they found out, they can be like "oh I didn't want this account created" or "I don't want to initiate HKN" and whatnot (a bit of data ethics precedence and it may scratch a bit on us using other people info without their permission with laws probably on that stuff based on precedence). My recommendation is to keep this feature and have students fill out a Google Form or something on the website to request the creation of the data OR have a preapproval process first (and I'm not talking about going to Admin and checking off everyone who is a Candidate ... that was annoying back in Spring 2021 during my officer time)
> 
I haven't settled on just how I'm envisioning this, mainly cause I want to have some sort of connection to CalNet, just not sure if it's possible to do what I have in mind for it. The pipeline will probably look like: during CM, VP tells all the candidates to make an account, RSVP to CM (an external event), and enter the attendance form password. As soon as they're confirmed, their account becomes a candidate account. This way they're introduced and taught about the website events pipeline at the same time, and it's as secure as having them fill out the Google form.
> Let alone I don't think such a reasonable, secure, and more legal replacement feature exists yet in your PR, so I'm pushing back against its deletion for now
> 
We're not provisioning candidates for another 3.5 months. It'll be done by then (probably within the next 3 days)
> Also, the VP or CSec confirm of challenges thing is there for a reason from the past (maybe to make sure the challenge meets the requirement of not being from a committee officer or from Byte or smth), but I'd consult VP and CSec if they are okay with this feature removed ... I'm neutral about this change as long as it goes through them
Good point, I'm gonna be talking to current VP (Shalin) in a few days, so will bring it up then. It's a 3 line change anyways, just adding back the field and redefining the confirmed and rejected properties.

PR #478, opened 2022-09-07
Revert "Tutoring scheduler and tutee view (#476)"

bri25yu:
Reverting to old functionality to retain usability of preference interface

This reverts commit 938ba3c2a5cbfd41d56931116563987b22ef1e07.


bri25yu:
The formatting checks that aren't passing are not due to this PR. They will be fixed in a subsequent PR

Issue #483, opened 2022-09-20
Using "conda" instead of Vagrant for M1 and Windows bugginess


ochan1:
Two reasons for this:
1. M1 MacBooks don't support Vagrant and VirtualBox, and solutions out there are either slow or paid
2. Windows Vagrant lately has been having issues with .venv folder linking permissions and output timeout that is annoying, and also the initialization of the VM box that takes minutes

A current solution for M1 MacBooks is https://github.com/compserv/hknweb/pull/431, but if conda is an all in one solution, that would be great ... and also good if it's a all in one for Windows too since the barrier to knowing if Linux packages or equivalent work on Windows is high and may be too manual

Originally, a Linux VM is used to emulate the production server as much as possible, but seeing as most of the work is on Python, any dependencies installed on Linux is rare. Many normal Python dependencies work on Windows and Linux operating systems anyway.

ALL of these are in conjunction required to get this to work:
- [x] Instructions to install conda. A huge requirement is to have a flag / setting checked during installation to NOT have conda as base environment (many of Oscar's friends who used conda in EECS 16A did not like it doing that). Backup instructions too where if they forgot how to deactivate (which isn't too bad anyway from Oscar's top of head)
- [x] Explore if conda is a good alternative -- `conda create -n hknweb python=3.7.3; conda activate hknweb;` then Makefile's make venv and make install, etc.
- [x] Need a way to run Makefile and support on Windows (Example of needed support: `make venv` uses "source .venv/bin/activate" while Windows is always ".venv\Script\activate" ... where "source" can be used on Git Bash terminal)

ochan1:
https://www.notion.so/hkn/Using-conda-instead-of-Vagrant-b92894a9bd7c4904b34d80696a46fb82

ochan1:
Completed
OCF and Dev uses Conda

See
- https://github.com/compserv/hknweb/pull/491
- https://github.com/compserv/hknweb/pull/503
- https://github.com/compserv/hknweb/pull/504

PR #487, opened 2022-09-23
Retire Vagrant for Development, direct to Machines + MySQL install on Prod ONLY

ochan1:
Two reasons for this:

1. M1 MacBooks don't support Vagrant and VirtualBox, and solutions out there are either slow or paid
2. Windows Vagrant lately has been having issues with .venv folder linking permissions and output timeout that is annoying, and also the initialization of the VM box that takes minutes

This also depreciates the need solution for M1 MacBooks is https://github.com/compserv/hknweb/pull/431 since it is identified MySQL isn't needed on Development, only Production

Originally, a Linux VM is used to emulate the production server as much as possible, but seeing as most of the work is on Python, any dependencies installed on Linux is rare. Many normal Python dependencies work on Windows and Linux operating systems anyway.

On Windows:
* Due to the Makefile syntax and the need for a temp variable (unless "set" is used, but too OS specific and will get clunky) and directory separators ("\" vs "/"), Git Bash is the standard terminal used.
* Make is installed via GNUWin32 by terminal from winget (easiest) and also Environment Variable assigned

This allows a lesser barrier to development to avoid possible hassles for development in exchange for more than likely minimal benefit from protection against dev vs. prod OS problems, since we rarely nowadays install OS specific libraries

Tested on Windows Git Bash and Windows WSL, untested on a fresh Unix or MacBook


ochan1:
Conda found to be not needed
Just install Python 3.7 and Makefile commands take care of the rest
It's either hand install Conda or Python 3.7, and Python 3.7 is a bit lighter and less "too much stuff"

The only reason why we may need it is if a library desperately needs OpenSSL on MacBook

In the meantime, should address this issue: https://github.com/compserv/hknweb/issues/483, especially with this comment's info on decision

ochan1:
Conda possibly may be needed actually for potential futureability support (e.g. allow an upgraded Python without relying on OCF)

Follow up on linked issue for more details

For now, closing PR

PR #491, opened 2022-09-27
Retire Vagrant for Conda + MySQL install on Prod ONLY

ochan1:
Two reasons for this:

1. M1 MacBooks don't support Vagrant and VirtualBox, and solutions out there are either slow or paid
2. Windows Vagrant lately has been having issues with .venv folder linking permissions and output timeout that is annoying, and also the initialization of the VM box that takes minutes
This also depreciates the need solution for M1 MacBooks is https://github.com/compserv/hknweb/pull/431 since it is identified MySQL isn't needed on Development, only Production

Originally, a Linux VM is used to emulate the production server as much as possible, but seeing as most of the work is on Python, any dependencies installed on Linux is rare. Many normal Python dependencies work on Windows and Linux operating systems anyway.

On Windows:

* Due to the Makefile syntax and the need for a temp variable (unless "set" is used, but too OS specific and will get clunky) and directory separators ("\\" vs "/"), Git Bash is the standard terminal used.
* Make is installed via GNUWin32 by terminal from winget (easiest) and also Environment Variable assigned
This allows a lesser barrier to development to avoid possible hassles for development in exchange for more than likely minimal benefit from protection against dev vs. prod OS problems, since we rarely nowadays install OS specific libraries

This PR also makes Conda the standard for development instead of using "venv" (conda and venv doesn't go together as well apparently), and retains Pip as the package manager installed directly into the Conda Environment, and the next deploy will use conda (as long as Miniconda is installed on the OCF server)

Brand new instructions on README
Once this is OK, I'll work on the "Wiki" as well

Tested on Windows Git Bash and Windows WSL, untested on a fresh Unix or MacBook


Makefile
```diff
@@ -17,42 +15,36 @@ dev:
...
+conda-scratch:
+	# Makes hknweb environment from scratch (overwriting old one if exists)
+	conda create -n $(CONDA_ENV) python=$(PYTHON_VERSION) -y
+	@echo "When developing, activate the HKNWeb Conda environment with 'conda activate $(CONDA_ENV)' so Python can access the installed dependencies."
+
+.PHONY: conda
+conda:
+	@# We don't use "-y" because if environment recrated, it will destroy and reinstall ... since it is an all "yes"
+	@#  Just confirm to the default options (option given is either: Yes install OR No remove and reinstall)
+	echo -e "\n" | conda create -n $(CONDA_ENV) python=$(PYTHON_VERSION)
```
girantinas, line 28: Don't need this -e here.

Makefile
```diff
@@ -17,42 +15,36 @@ dev:
 livereload:
 	$(MANAGE) livereload $(IP):$(PORT)
 
-.PHONY: venv
-venv:
-	python3 -m venv $(VENV)
-	@echo "When developing, activate the virtualenv with 'source .venv/bin/activate' so Python can access the installed dependencies."
+.PHONY: conda-scratch
```
girantinas, line 18: Not sure why this is needed, the default option is all yeses anyways? The other command works fine for me, but maybe it breaks on a fresh install of conda?

Makefile
```diff
@@ -1,9 +1,7 @@
-PIP_HOME = $(shell python3 -c "import site; import os; print(os.path.join(site.USER_BASE, 'bin'))" \
-	|| python -c "import site; import os; print(os.path.join(site.USER_BASE, 'bin'))")
-
-VENV := .venv
-BIN := $(VENV)/bin
-PYTHON := $(BIN)/python
+CONDA_ENV := hknweb
+PYTHON_VERSION := 3.7.3
+# Require Python be ran insie Conda, otherwise directory error out
+PYTHON := "$(CONDA_PREFIX)"/python
```
girantinas, line 4: My conda python seems to be at "$(CONDA_PREFIX)"/bin/python

girantinas (changes requested):
After making the changes that I mentioned, I walked through the steps of the README and everything seemed to work fine. Great work!

Makefile
ochan1: Needed to confirm install on first time, but if the environment exists, it will destroy the old one and recreate

The default option respectively is Yes Install, but No Reinstall

"-e" also allows for the use of \n as newline instead of it as separate characters (like "\" and "n")

Makefile
ochan1: As noted in other command, plan "-y" says yes to everything

So it will do Yes Install, and also Yes (fresh scratch) Reinstall for someone who isn't careful

Makefile
bri25yu: What about `if [ conda env list | grep $CONDA_ENV | $(wc -l) -lt 1 ] then conda create ... fi`? We can check if the env by that name already exists

Makefile
bri25yu: See https://github.com/compserv/hknweb/pull/491#discussion_r990499490

Makefile
bri25yu: Pretty sure when you `conda activate ...` it'll set the env var for python to point to the correct python, but I don't have a link for this. A test is `which python; conda activate ...; which python`.

Makefile
ochan1: Conda Create checks if exists already, and we just use the default option of **deny** the overwrite with a simple new line

While the solution proposed is the alternate and does the similar behavior, next level consideration imo is what's simpler for future CompServ to read, and the current line is easier imo

Makefile
ochan1: Yes it does
Otherwise, it is blank
"/python" usually doesn't exist unless someone on purpose tried to make one

Following similar design decision and precedence of using venv Python, we try to target the hknweb Conda Python to avoid using another Python by accident

Yes, "which python" does show the Python path, but if not in conda environment, it uses a non-conda Python (unless they check the box to use Conda Python by default)

Makefile
bri25yu: AgreeGe, makes sense

Makefile
ochan1: It's possible to write a "make check-conda" and copy paste it on each Makefile command

I will investigate that as well

ochan1: Conda sanity checks pushed via commit ee219db

girantinas (approved):

ochan1:
Merging in because of checks from another PR, and need to test for OCF

PR #510, opened 2022-11-08
first attempt

nmacasaet1003:
finished checkpoint 1, hope this works


lo-maxwell:
I don't see any changes here, did you make sure to add/commit/push your work?

nmacasaet1003:
> I don't see any changes here, did you make sure to add/commit/push your work?

I swear I did, but knowing myself, I probably did something wrong. When I tried to push, it said something about a submodule or something like that. At this point, what should I do? Should I push again, or just push the final checkpoint, since I've made some progress since pushing the first checkpoint. If you'd like me to re-push, could I PLEASE get step-by-step instructions on how to push/make a pull request? I'm getting very frustrated with everything, and also very stressed about whether I'll be able to initiate.

lo-maxwell:
Looks good, good job! Some totally optional things to add if you still want to experiment with django: try making a link to/from the main site to the polls app (instead of having to go through the url), or changing the style of the app to match the rest of the website.

Also the problem with not being able to run createsuperuser was a result of the move to conda and *should* be fixed, so I think you should be able to go through the admin panel and create questions/choices for your app even without the secret key now as long as you're running in dev mode.

PR #511, opened 2022-11-11
Finish Part 6 Django Tutorial

luisCermeno:
My tutorial code can be found under hknweb/mysite.


lo-maxwell:
Tutorial looks pretty good, make sure to add `admin.site.register(Choice)` to `hknweb/polls/admin.py`. Also check the update I made in the compserv slack channel, we want the polls app to be an app within hknweb instead of being just a separate folder thrown in. 
Keep up the good work!

luisCermeno:
Thanks for the feedback Maxwell! Will apply the changes

PR #512, opened 2022-11-12
django tutorial finished

jhuey21:


lo-maxwell:
Looks like you're still working on this, but you're on the right track. Hopefully the setup/installation process wasn't too bad. Let us know if you feel stuck/need help with anything!

I put an update in the compserv channel that basically says you'll need to move the polls app into the actual hknweb site at some point, but it might be easier to finish the tutorial as a standalone segment first and then move it in afterwards.

jhuey21:
Finished part 7

lo-maxwell:
Looks good! Not much to say here, everything seems fine :)
If you want to continue experimenting with django, some optional things to do are: 1) modify the frontend to match the layout of the rest of the hknweb site, 2) make clickable links going to/from the main site to your polls app so the user doesn't have to go through the url, 3) play around with the admin panel and see how data is modified and where it goes (may require pulling the most recent version of the repo for this to work).

PR #513, opened 2022-11-14
Tutorial part 3 in progress

adymchenko:


lo-maxwell:
Your code compiles and I can run the website, so that's a good start! It looks like you skipped over some sections in the admin file (part 2 of the tutorial) - I don't see a way to add choices to questions. 

For the final submission, make sure to put the polls app inside the hknweb site as an additional app, so that we can view both the regular site and your created app from one server instance. (Take a look at my post in the compserv channel for details, or feel free to ask a question/dm an officer if you need help!)

Keep up the good work!

PR #515, opened 2022-11-14
Tutorial Part 2 Done, Part 3 In Progress

ventimdg:


lo-maxwell:
Hey there, it looks like you're missing the choices section of the admin panel. Might want to go back to part 2 of the tutorial and make sure you have the ability to add choices to your questions. Other than that, your code compiles and the website runs, so good job so far!

For your final submission we want the polls app to be an app within the hknweb site, rather than a standalone site. This means we should be able to run one server instance that can go to both dev-hkn and your created polls app. Take a look at the compserv channel in slack for more details. 

Make sure to ask questions and/or dm an officer if you need help! Or just to talk, we don't bite.

lo-maxwell:
Looks pretty good, just remember to remove extra comments like the ones in polls/views.py and common.py if they're just unused code. 
If you want to continue experimenting with django, some other optional things to try are 1) customizing the frontend to match the layout of the rest of the hknweb site, 2) making a link to/from the main site to your polls app so users don't have to access it only through the url, 3) messing around with the admin panel (and seeing how apps besides your polls app use the admin panel to modify data), though not sure if this works without updating to the newest version of the repo.

Congrats on finishing the candidate project! Closing this pr now.

PR #518, opened 2022-11-21
SavitBhatCheckpoint

savitb:
part 1 completed part 2 in progress


lo-maxwell:
Good job getting the pr working, looks like you're on the right track. When you finish the tutorial and move it as an app inside the hknweb site, you might not need to have the virtual environment as we already have one set up through conda. 

As a reminder the entire project is due 12/2, let us know if you're running into any other problems!

lo-maxwell:
So we actually want the polls app to be an app within the hknweb site, rather than just throwing the entire mysite folder in (as it can't be accessed from the site right now). See if you can modify this to have your polls site as an additional app.

In the end product we won't need any of the stuff outside the polls folder (so no mysite/venv stuff), but you will need to modify the settings.py and urls.py files in the common folder so that it recognizes /polls as an app.

lo-maxwell:
Looks good. 
If you want to continue experimenting with django, some optional things to do are: 1) modify the layout of the polls app to match the rest of the hkn site, 2) make links going to/from the main site to the polls app, 3) mess around with the admin panel and figure out how data is modified.

PR #527, opened 2022-12-01
Candidate Project: Done with Part 7 

luisCermeno:
Hi Compserv, I finished part 7 and also included `polls` as an hknweb app. Thank you for the feedback last time!


lo-maxwell:
Nice job! Next time let's remove the commented out code from polls/urls.py and polls/views.py as it no longer serves a purpose after we upgraded it to the better version. 
If you want to continue experimenting with django, some optional things to do are: 1) modify the frontend to match the layout of the rest of the hknweb site, 2) make links going to/from the main site to your polls app so the user doesn't have to manually enter the new url, 3) mess around with the admin panel and see how data is modified on the main site (may or may not work unless you update the repo to the newest version). 

PR #528, opened 2022-12-02
Added polls app for candidate project

WillG777:
Not 100% sure if this works, wasn't able to test it yet (not sure which venv I should be running commands in). The polls app itself should be fine though; I initially had it in a separate site as the tutorial did and successfully tested that site before moving polls to hknweb.


lo-maxwell:
So this is on the right track, but you'll also need to add the app to the settings.py and urls.py files in /common (so that the site recognizes it as an official app).

WillG777:
Thanks for the info! Added those, seems to work now. (There are some failed test cases when I run `coverage run`, which I think have to do with string representation bugs. Not sure if they affect anything; will try to debug tomorrow.)

lo-maxwell:
Looks good! I seem to remember it not working with the references to polls in settings/urls being so far down, so if it doesn't run properly try moving them up where the rest of the apps are, but if it works then that's good. 
If you want to continue experimenting with django, some optional things to do are: 1) modify the layout of your polls app so that it matches the rest of the hkn site, 2) make links going to/from the main site to the polls app, 3) mess with the admin panel and see how data is modified and where it goes on the site (may or may not work without updating to the newest repo). 

PR #530, opened 2022-12-03
Franz K

FranzKieviet:
My PR!


lo-maxwell:
Looks good, for a real merge I would want the extra mysite/ files deleted since they don't do anything but the hknweb part of the app works. 
If you want to continue experimenting with django some optional things to do are: 1) modify the layout of your polls app to match the rest of the hknweb site, 2) make links going to/from the main site to the polls app so the user doesn't have to directly change the url, 3) mess around with the admin panel and see how modifying data affects the site.

PR #533, opened 2023-03-02
Add end time None check

ochan1:
The fact that it is required, the form validation will error the entry if the end time is forgotten, instead of it simply erroring out


ochan1:
Wait, never mind
There is an end_time value according to the email

PR #534, opened 2023-03-02
Add start_time and end_time None check due to time input misformat

ochan1:
Formatting of the times become wrong and thus become None

The error message will recommend to use the time picker

Sample error log
start_time = '03/11/2023 2:00 PM'
end_time = '03/11/2023 4:00PM'

Notice the End Time format is not with the Start Time format standard, and thus is None


508312 (approved):
lgtm

PR #538, opened 2023-04-07
Compserv Candidate Project Checkpoint and Checkoff

williamT386:
Completed Django tutorial 1-7 and added the `polls` app to the route `polls/`.


lo-maxwell:
Not sure what happened here but you somehow got the entire commit history from 2 years ago into your PR. Since we aren't actually merging this in and I can find the relevant files it's fine for the candidate project, but if this happens in an actual project you should probably fix it before submitting a pr. Having all these commits makes it impossible to tell what your changes were (as opposed to someone else's 2 year old changes). 
With that said, congrats on finishing the candidate project! Some totally optional things to do to expand on this could be adding links/buttons to the actual webpage to go between the main site and your polls app (instead of having the user put /polls in the url), or changing the layout to match the rest of the site. 
Closing this PR now since you have also finished the django quiz.

PR #540, opened 2023-04-10
checkpoint (django tutorial pt 1)

0315jaewon:


lo-maxwell:
Looks good so far, keep up the good work! Remember that for the final submission you'll have to integrate your polls app with the hknweb site rather than just slapping the files in, but that's something you can deal with after finishing the tutorial. 
Also take a look at the django quiz, it has some supplementary material and is meant to be worked on during/right after finishing the tutorial.

PR #542, opened 2023-04-10
candidate project checkpoint submission, completed up to pt 4.

michaelwiradharma:
whoops this might be a second pull request lol


lo-maxwell:
Looks like you put in 2 copies of /mysite, one in and one outside of hknweb. In the future we only need 1 copy :)
Also remember that for the final submission we want the polls app to be integrated with hknweb rather than just having the files slapped on, but that's something to think about after finishing the tutorial. 
Keep up the good work, and take a look at the django quiz if you haven't yet! There's some supplementary material there that might help during/after finishing the tutorial.

michaelwiradharma:
Candidate project final submission

lo-maxwell:
Looks good! If you want to expand on this, some totally optional things to do are adding the polls app to the admin panel, creating clickable links going from the hknweb site to your app and back, or changing the layout to match the rest of the site. 

Congrats on finishing your candidate project! Since you've also completed the quiz, I'll close this PR now.

PR #544, opened 2023-04-11
checkpoint

alex-chen1:
checkpoint progress in /checkpoint


lo-maxwell:
Looks good so far! Remember that for the final submission you'll have to integrate the polls app with the hknweb site, but that's something to deal with after finishing the tutorial. 
Keep up the good work and let us know if you need anything!

PR #547, opened 2023-04-25
finished candidate project

JustinH0404:
finished candidate project


lo-maxwell:
It doesn't look like your polls app is in the right spot. It should be within the first level hknweb (same directory level as academics or candidate). You'll also need to update the urls.py file for hknweb. 

Also, since you haven't attended a weekly meeting yet, I expect to see you there for tomorrow's meeting. Let me know asap if you can't make it!

PR #550, opened 2023-04-26
Final Candidate Project

mannymnz:
completed tutorial is unser "/polls/"


lo-maxwell:
Make sure to also update hknweb/settings/common.py with the polls app. Also, if this were an actual merge, changes you make would be applied to the production site. Changing debug mode in prod.py is fine for testing, but don't keep it in your final code.

Issue #537, opened 2023-04-06
Phantom event on google calendar


lo-maxwell:
If a user adds the event calendar to their personal google calendar and then an event date is changed, it is not deleted from their personal calendar (a second event will appear with the correct time that can be rsvped to, but the old event is just floating there and cannot be interacted with). 

"Hey y'all, is there a way to have events removed from google calendar if they're removed from the hkn calendar? I have a few "phantom events" that were deleted off the hkn calendar but I can't get them off my calendar"

lo-maxwell:
Also if someone uses emojis in the event name it creates the event fine in google calendar but crashes dev-hkn, which can cause buildup of multiple bad events on google calendar. Look into doing all or nothing for creation of events (or do dev-hkn first, then gCal, then rollback if gCal fails)

bri25yu:
From @oliver-ni: hknweb uses OCF's mysql database server which uses utf8mb4 encoding which is 4 bytes and supports emojis. So the database encoding itself may not be the problem, maybe it's the database encoding setting on the Django side https://github.com/compserv/hknweb/blob/master/hknweb/settings/common.py#L106

oliver-ni:
Emoji issue has been fixed in #551

PR #551, opened 2023-09-02
Set charset for actual unicode support

oliver-ni:
This PR fixes unicode support, allowing for (among other things) emoji to be used in event names.

Tested and ready to go.


oliver-ni:
Was mentioned here: https://github.com/compserv/hknweb/issues/537#issuecomment-1516787187

lo-maxwell (approved):
Tested and works fine locally 👌

ochan1:
This looks good, I would have put it here but didn't know how to test or where to put it

Please make sure this works on Production, since sqlite3 dev mode already is fine with emojis

PR #558, opened 2023-10-05
Expose personal calendar in iCal format instead of provisioning Google Calendars

oliver-ni:
[relevant slack thread](https://hkn.slack.com/archives/C159XC064/p1696479787822419)

Instead of provisioning and maintaining a Google Calendar for each person, we have a route on dev-hkn that just outputs the rsvp’d events in the standard iCal format that can be parsed by any calendar application. Google Calendars that were already provisioned will continue to work and be updated. However, they will no longer be provisioned for new candidates.

This has already been tested & a deployed has been made (would normally wait but it's urgent for CM tomorrow). You can see my account has both the old link and the new link here.

CleanShot 2023-10-05 at 00 05 39@2x


oliver-ni:
CleanShot 2023-10-05 at 00 13 27@2x

PR #560, opened 2023-10-22
Fix for calendar overflow issue.

508312:
If screen size was above 1000px the offset of 12.5% from the left would be added to the calendar. However this would cause calendar with sidebar to overflow past the page. Changed the offset of calendar with sidebar to be 3.33%, this will mask the issue, until proper fix is found. (original calendar without the sidebar has the overflow issue too, but because of how far it is from the border it was never noticed)

### DIFF
--- a/hknweb/static/css/base.css
+++ b/hknweb/static/css/base.css
@@ -24,6 +24,7 @@ body {
 
 .content {
     padding-bottom: 1em;
+	overflow:auto;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


508312:
Sorry for excessive commits, don't know if there is a way to squash them inside the pull request, if this is to be pulled just squashing to 1 commit should be pretty.

lo-maxwell:
Sorry for the delay, this looks good, tested locally (by copy pasting changes lol) and everything seems okay. I'll merge this in when I get the chance.

Issue #495, opened 2022-10-05
Revisit model field `max_length` properties


bri25yu:
From anson.tiong:
> Was there a reason for this limit? I'd like to give more flavortext for my event, but I'm limited to these 500 characters now

![image (1)](https://user-images.githubusercontent.com/46059916/194166090-aa3d2ed7-d9fb-4040-a7ff-b1014cf9dc33.png)

This specific example is for https://github.com/compserv/hknweb/blob/master/hknweb/events/models/event.py#L20, but we should revisit all the current restrictions on `max_length` on all our models even if this is a pretty extreme scenario.

PR #561, opened 2023-11-05
Update max character count on fields - Owen Issue #495

owen2608:
I changed the limit from 500 to 2000 in 4 cases, mainly being the texts of various fields


FranzKieviet:
Looks great! Thank you Owen!

PR #567, opened 2023-11-29
Update Candidate Account Provision to Use Bulk Creation

PBales1:
Previous implemented used for loop to create accounts when provisioning candidate accounts via csv input, which was slow.  This change removes the for loop when creating an account and processing it into the db, instead bulk provisioning and processing accounts into the db.  Video attached below demonstrates process; note the fast processing speed compared to current implementation (Video size too large, so placed in linked drive folder).

https://drive.google.com/file/d/1IipyCfeK6iFGA98kBw9gvlkXljmFcEnR/view?usp=sharing


hknweb/events/migrations/0013_alter_icalview_options.py
oliver-ni, line 1: This migration is probably not necessary

hknweb/forms.py
```diff
@@ -236,30 +236,26 @@ def generate_password() -> str:
...
                 username=row["username"],
                 first_name=row["First name"],
                 last_name=row["Last name"],
                 email=row["Berkeley email"],
-                password=password,
-            )
-            user.save()
-
-            # Add user to the candidates group
+                password=generate_password(),
```
oliver-ni, line 249: Password needs to be hashed. `create_user` handles this, but without it we need to handle it ourselves

oliver-ni (changes requested):

hknweb/forms.py
oliver-ni: Probably relevant: https://docs.djangoproject.com/en/4.2/topics/auth/passwords/#django.contrib.auth.hashers.make_password

PBales1: Sounds good; is it a salted hash by any chance or just a standard hash?

oliver-ni: On second look, there's also https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.models.User.set_password which might be better here

oliver-ni: (with make_password I think passing in None for the salt makes it use the default option which is probably fine, not sure what we're running in production)

PBales1: Sounds good.  I'm guessing that the password hashing algorithm uses an intentionally slow algorithm so it slows things down a bit now, but the process is still a bit faster than before.

PBales1: Alright, should be good now.  Curious if bulk_create() also hashes the password for you (which would make set_password() unnecessary), but probably not since the overhead with the PBKDF alg slowing things down is just always gonna be there unfortunately.

hknweb/forms.py
```diff
@@ -236,30 +236,30 @@ def generate_password() -> str:
...
+        users = User.objects.bulk_create(user_data) # Bulk process accounts into database
+        
+        # Save each user, add to candidate group set, and add information for emails
+        email_information = []
+        for user in users:
             group.user_set.add(user)
-
-            # Add information for sending emails
-            email_information.append((user, password))
+            email_information.append((user, user.password))
```
oliver-ni, line 262: Thanks! Now this part needs a change too because user.password is the hashed one

oliver-ni (changes requested):

hknweb/forms.py
```diff
@@ -236,30 +236,30 @@ def generate_password() -> str:
...
+            password = generate_password()
+            user.set_password(password)
+            user_data.append(user)
+        
+        users = User.objects.bulk_create(user_data) # Bulk process accounts into database
+        
+        # Save each user, add to candidate group set, and add information for emails
+        email_information = []
+        for user in users:
             group.user_set.add(user)
```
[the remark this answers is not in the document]
oliver-ni: But yeah I think password hashing is probably the slower operation

oliver-ni (approved):

oliver-ni (changes requested):
Wait jk we also need to create Profile objects since signals aren't called on bulk writes

PR #568, opened 2023-12-02
added page for dynamically changing sponsors at /indrel_admin

Hoponga:
title basically describes the change 

Can add sponsor name, URL, and a picture, saves to /static/img/sponsors 

I keep track of the current sponsors with a metadata file, where each line contains name, URL, and filename for the company. every time someone submits the form, it updates the metadata file to either add new companies or remove some 

if someone just deletes the entry for a company on the admin page, it deletes that sponsor from the metadata file, & if a file is not added when submitting on the admin page, then it doesn't change the file that's already in the sponsors folder 

it's low-key kinda jank with the metadata file but i didn't wanna change anything with the django database & have to run migrations and stuff, so i just did it this way


ochan1:
Could we use the [FileField](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#django.forms.FileField) to upload images of sponsors? Why does it have to be static?

PR #573, opened 2024-03-16
Onboarding Exercise

joshualu777:
Finished onboarding exercise. Created point of contact field.


joshualu777:
- Tested candidate creation process to make sure that it is working, able to add more than 10 names per CSV file
- Improved duplicate username checking to ignore usernames that appear more than once in CSV file
- Added duplicate email checking, website no longer crashes when duplicate emails appear in CSV file
- Added case-insensitive checking for emails, emails that already exist will be skipped during account creation
- Wrote statement to print all skipped accounts in the CSV file

PR #578, opened 2025-02-07
Remove resume review page

ydoccr:


508312:
Whoops I left a space between <p> tag and it's body. Someone might find it ugly, but it looks fine to me, I will merge it in :D. (google html formatting guide did not specify if this should not be done so idk lol)

PR #581, opened 2025-03-16
Search Bar And Current Officer Filter For Leaderboard

ryantrinh05:
Added 2 separate filters onto the user data that gets fed into the paginator that displays the leaderboard.

The first one compares what is typed in the search bar and the user's data (username + name) and filters the users accordingly. The other one filters based on "current officers" which is determined if a user is apart of a committeeship. Currently it only checks for Spring 2025 committeeships and doesn't automatically update. It also considers officers, aos, and committee members to all be current officers. 

No Filters
nofilter

Current Officer Filter only
currofficersonly

Search Filter Only
johnfilter

Both Filters
johnandofficer

No Matching Users
none


hknweb/events/views/aggregate_displays/leaderboard.py
```diff
@@ -28,8 +28,8 @@ def get_leaderboard(request):
     if active:
         current_semester = Semester.get_current_semester()
         curr_committeeships = Committeeship.objects.filter(
-            Q(election__semester__semester=current_semester.semester) &
-            Q(election__semester__year=current_semester.year)
+            Q(election__semester__semester=current_semester.semester)
```
508312, line 31: Bruh, in my opinion this is uglier....

508312 (approved):
Looks good to me. Potentially some variables will be more clear if renamed, such as query, or active(there, in theory, can be active people who are not members of a committee, but I do not know the official terms, and active sounds clear to me) Have no better names off the top of my head.

hknweb/coursesemester/models.py
```diff
@@ -37,3 +37,21 @@ class Semester(models.Model):
...
+        if latest_fall.exists():
+            return latest_fall.first()
+
+        # Add summer?
+        latest_spring = latest_year_semesters.filter(semester="Spring")
+        if latest_spring.exists():
+            return latest_spring.first()
+
+        # Should never get here.
+        return latest_year_semesters.first()
```
508312, line 57: This may be overly complicated. However, I did not want to just do Semester.objects.order_by("-year", "semester").first(), since sorting strings doesn't seem like the most error free way to determine final semester. That being said this function will still need to be rewritten if internal format changes, so maybe simpler solution will be clearer

PR #583, opened 2025-03-30
Use separate lines for first/last name and accommodate berkeley.edu emails

spiralsim:
Screenshot 2025-03-30 at 11 24 22 AM


508312 (approved):
Looks good. Alternative solution would be to replace "@" sign with " [at] " or something similar, but I think this will do.

PR #591, opened 2025-05-02
Course descriptions

ryantrinh05:
Built From Prav's Course Descriptions

- Markdown instead of textfields
- Committee Portal for Tutoring
- Course Description Page for Tutoring
- Ways to edit descriptions directly on the page for tutoring


508312 (approved):
Looks good, but one thing to continue is to maybe structure classes like old website did. This will work but might not be the most robust.

PR #592, opened 2025-05-31
Bitbyte tree implementation

508312:
Rough implementation of bitbyte tree.
Base data is currently in csv under static, its 20kb so surely thats not a problem to keep it there.


ryantrinh05 (approved):
All looks good. 

- Future updates to the bitbyte graph can be done by candidates

